PR #4, opened 2026-04-02
docs: correct link to beads repo

trythrow:


bdfinst:
After reviewing the utility of beads with the current version of Claude, I've removed the dependency. Thank you so much for the help, though!

PR #541, opened 2026-07-01
Docs/skills by capability

bdfinst:


bdfinst:
Closing as superseded. Comparing `docs/skills-by-capability` against `main` produces **zero file changes** (4 commits ahead, 0 net diff) — the substantive commit `docs: group the skills catalog by capability, not invocation type` has already landed via subsequent merges of main into this branch, so there is nothing left to merge. Only the title-lint check was failing; renaming the title would have satisfied lint on an empty PR without adding value.

Issue #859, opened 2026-07-05
feat: trajectory-level telemetry — boundary-events channel and intervention events


bdfinst:
**Classification**: Weaker · **Layer**: Hook + Knowledge · **Priority**: High (rank 1)

**Paper recommendation (§3.5.1, §5.2.1)**: record the decision process — hook blocks/permission denials, human interventions, per-agent step events — link signals to artifacts so runs are replayable, and attribute failures to specific harness components. Three complementary channels: eval scores (task-level), traces (trajectory-level), policy-gateway events (boundary-level). Industry-wide, step-level failure attribution is only 14–53% accurate without structured traces.

**What we have**: `hooks/telemetry.py` records `{ts, event, name, outcome}` grammar-matched events plus artifact-usage counts; `cost_meter.py` records per-agent token/cost. The boundary channel (destructive-guard warns/blocks, verify-guard blocks, pre-commit-gate details) and intervention events (override/pause/stop) are not captured, so `/session-review` and `/harness-audit` reason from outcome counts without causes.

**Proposed**:
- Every hook block/warn/bypass writes `{ts, hook, tool, decision, matched_rule}` to `metrics/boundary-events.jsonl`; intervention keywords logged as events.
- Stabilize all metrics JSONL schemas in a new `knowledge/telemetry-schema.md` so session-analysis and cross-machine aggregation compose.
- Keep the existing "never free text" privacy stance — record rule IDs, not command text.

**Dependencies**: session-review, harness-audit, session-analysis agent, cost-report. **Complexity**: Medium.

This is the foundation gap: eval-gated mutation, lesson validation, and ablation (sibling issues) all consume this data.

Source: competitive analysis vs *Code as Agent Harness* (arXiv 2605.18747), full report at `reports/competitive-analysis-2026-07-05.md` (local, session of 2026-07-05).

---

### Intent Description

[...]

PR #890, opened 2026-07-06
feat: context-sensitive destructive_guard — branch/remote-aware risk escalation

### LINKED ISSUE #862: feat: context-sensitive destructive_guard — branch/remote-aware risk escalation
**Classification**: Weaker · **Layer**: Hook · **Priority**: High, quick win (rank 3)

**Paper recommendation (§5.2.5)**: permission decisions must depend on arguments and environment state, not command identity — the same `git push --force` is routine on a feature branch and catastrophic on `main`.

**What we have**: `hooks/destructive_guard.py` lowercases the command and substring-matches pattern lists in `destructive-commands.json`; the only state consulted is the careful-mode boolean. Branch, remote, and environment are never inspected — force-push to `main` warns identically to a feature branch.

**Proposed**:
- Cheap context probes in the hook (current branch vs default branch, remote URL, `CI` env), stdlib-only.
- Per-pattern risk-escalation rules in the existing JSON format — e.g. force-push/reset/branch-delete patterns escalate warn→block when the target resolves to the default branch, even outside careful mode.
- Log decisions to the boundary-events channel (see the trajectory-telemetry issue).

**Dependencies**: careful/guard skills, destructive-commands.json. **Complexity**: Small.

Source: competitive analysis vs *Code as Agent Harness* (arXiv 2605.18747), full report at `reports/competitive-analysis-2026-07-05.md` (local, session of 2026-07-05).

---

### Intent Description

`hooks/destructive_guard.py` currently treats every match of a destructive pattern identically: warn (exit 0) by default, block (exit 2) only when careful mode is on. The risk of the same command varies enormously with context — `git push --force origin feature/wip` is routine, while `git push --force origin main` can destroy shared history — yet today both produce the identical CAUTION line. The guard consults exactly one piece of state (the careful-mode boolean) and never inspects the branch, the remote, or the environment.

This change makes the guard context-sensitive. When a destructive git pattern matches, the hook runs cheap, stdlib-only context probes (current branch, resolved default branch, remote URL, `CI` env) and evaluates per-pattern escalation rules declared in `destructive-commands.json`. When a rule's conditions match with high confidence — e.g. a force-push, reset, or branch-delete whose target resolves to the repository's default branch — the action escalates from warn to block **even when careful mode is off**. Everything else behaves exactly as today: patterns without escalation rules, non-git commands, and any environment where the probes cannot run (no git repo, git absent, subprocess timeout) fall back to the current warn/careful behavior.

Each escalation decision (matched pattern, probe results, chosen action) is additionally logged to the boundary-events channel introduced by #859 when that channel exists — a soft dependency; the guard is fully functional without it.

### Architecture Specification

**Components**

- `plugins/dev-team/hooks/destructive_guard.py` — extended with:
  - *Context probes* (module-level, stdlib-only): `_current_branch()` (`git rev-parse --abbrev-ref HEAD`), `_default_branch()` (`git symbolic-ref --short refs/remotes/origin/HEAD`, falling back to a `main`/`master` local-branch existence heuristic), `_remote_url()` (`git remote get-url origin`), and a `CI` env check (`os.environ`). Each git probe is a single `subprocess.run` with a short timeout (~1 s) and `Optional[str]` return — `None` on any failure (not a repo, git missing, timeout, non-zero exit).
  - *Escalation evaluator*: after a pattern match and only for matched commands (lazy — no probes run when nothing matches), look up escalation rules for the matched pattern, evaluate conditions against the command text + probe results, and escalate warn→block on a high-confidence match. Careful-mode block behavior is unchanged and takes precedence (already blocks everything matched).
- `plugins/dev-team/hooks/destructive-commands.json` — schema extension: a new optional top-level `"escalations"` key. Existing keys (`file_destruction`, …, `safe_allowlist`) keep their array-of-strings shape untouched.
- Boundary-events emission (soft dependency on #859): if the channel exists (emitter module under `hooks/lib/` or its JSONL sink, per #859's landed shape), append one decision event `{ts, hook: "destructive_guard", pattern, category, action: warn|block, escalated: bool, reason, probes: {current_branch, default_branch, remote_url, ci}}`. If absent, skip silently — never fail, never print.

**Interfaces — escalation rule shape**

```json
{
  "escalations": [
    {
      "pattern": "git push --force",
      "base_action": "warn",
      "escalate_to": "block",
      "when": { "target_branch": "default" }
    },
    {
      "pattern": "git branch -D",
      "escalate_to": "block",
      "when": { "target_branch": "default" }
    },
    {
      "pattern": "git reset --hard",
      "escalate_to": "block",
      "when": { "current_branch": "default" }
    }
  ]
}
```

- `pattern` — must equal an entry already present in one of the pattern arrays (rules for unknown patterns are ignored).
- `base_action` — optional, informational; defaults to `"warn"` (today's behavior). The only supported escalation today is warn→block.
- `when` — conjunction of condition keys. Initial vocabulary:
  - `target_branch: "default"` — the command text explicitly names the resolved default branch as its target (e.g. `git push --force origin main`, `git branch -D main`), **or** the command targets HEAD implicitly (bare `git push --force`) and `_current_branch()` == `_default_branch()`.
  - `current_branch: "default"` — `_current_branch()` == `_default_branch()` (for HEAD-local commands like `git reset --hard`).
  - Probe outputs `remote_url` and `ci` are captured for logging and reserved as future condition keys; no shipped rule uses them (see Ambiguity Log).
- Any condition that cannot be evaluated (probe returned `None`) evaluates **false** — no escalation, current behavior preserved.

**Dependencies**

- Careful state (`hooks/careful-state.json` via `/careful`, `/guard`) — read as today; careful block precedes escalation logic. Escalated blocks emit a distinct message (naming the branch context, not careful mode) so the "/careful off" hint is not shown for a block /careful off would not lift.
- #859 boundary-events channel — **soft**: guard must work, and all acceptance criteria except the logging one must pass, on a tree where #859 has not landed.

**Constraints**

- Python 3.8+ stdlib only (ADR 0014/0015). No new files required beyond edits to the hook, JSON, and tests.
- Probes are fast and fail-safe: one `subprocess` call each, short timeout, run only after a destructive match; guaranteed fallback to today's behavior when git/repo/timeout fails. Non-matching commands incur zero added cost.
- False-positive budget: escalation fires only on high-confidence evidence — the resolved default-branch name appearing as the command's explicit target, or the resolved HEAD equaling the default branch for HEAD-scoped commands. Ambiguity (unresolvable default branch, unparseable target) means no escalation.
- Backward compatibility: `destructive-commands.json` without an `escalations` key — and the inline fallback defaults — behave byte-for-byte as today (same messages, same exit codes). Existing tests in `plugins/dev-team/tests/hooks/test_destructive_guard.py` must pass unmodified.

### Acceptance Criteria

Each criterion is named for what it asserts; pass condition follows.

- **force_push_to_default_branch_blocks_without_careful_mode** — with careful mode off, current branch `main`, default branch resolved to `main`, `git push --force origin main` (and bare `git push --force`) exits 2 with a block message that names the default-branch escalation (not careful mode).
- **force_push_on_feature_branch_keeps_warn_behavior** — with careful mode off on a feature branch, `git push --force origin feature/x` exits 0 with today's exact CAUTION output.
- **branch_delete_of_default_branch_blocks** — `git branch -D main` (default branch resolved to `main`) exits 2 regardless of the currently checked-out branch.
- **reset_hard_escalates_only_when_head_is_default_branch** — `git reset --hard` exits 2 when the current branch is the default branch, and warns (exit 0) as today on a feature branch.
- **guard_degrades_outside_git_repo_or_on_git_failure** — when the probes fail (no git repo, git binary absent, subprocess timeout), every matched command produces today's warn/careful behavior; the hook never raises and never exceeds the probe timeout budget.
- **json_without_escalations_key_is_byte_compatible** — a config with no `escalations` key (and the inline fallback path) yields identical stdout and exit codes to the pre-change hook for the full pattern set; existing tests pass unmodified.
- **careful_mode_block_still_covers_everything** — with careful mode on, all matched commands block exactly as today, escalation rules or not.
- **each_probe_and_escalation_path_has_pytest_coverage** — `plugins/dev-team/tests/hooks/test_destructive_guard.py` gains cases for each probe (success, non-repo, timeout/absence) and each escalation path above, using the existing monkeypatch/capsys conventions; probes are faked at the subprocess seam so tests stay deterministic and offline.
- **decision_logged_when_boundary_events_channel_exists** — when the #859 channel is present, one event per guard decision (warn, block, escalated-block) is appended with pattern, action, escalation flag, and probe snapshot; when absent, no error and no extra output.

### Ambiguity Log

| Decision | Classification | Resolved By | Rationale / Answer |
|----------|---------------|-------------|-------------------|
| How the default branch is determined | `inferable` | inference | `git symbolic-ref --short refs/remotes/origin/HEAD` is the authoritative cheap probe; when unset (common on fresh clones), fall back to whichever of `main`/`master` exists locally; if both or neither resolve unambiguously, treat as unresolved → no escalation. Matches the false-positive budget: escalate only on high confidence. |
| Should a context-escalated block be overridable, and how? | `requires-stakeholder-input` | human (2026-07-05) | ONE-SHOT ENV VAR — e.g. `DEV_TEAM_GUARD_OVERRIDE=1` on the single command; explicit in the transcript, auto-expires with the command, and is logged as a bypass boundary event. |
| Does branch-delete of the default branch escalate too? | `inferable` | inference | Yes — the issue text lists force-push/reset/branch-delete together, and `git branch -D <default>` names its target explicitly in the command, making it the highest-confidence match available. |
| Does `git reset --hard` escalate on the default branch (local HEAD, different risk)? | `inferable` | inference | Yes, but keyed on `current_branch == default` rather than a command-text target: it destroys local state only (reflog-recoverable), so it is lower blast-radius than force-push, yet the issue names it and discarding local state while on the default branch is the exact context-mismatch this feature exists to catch. Reset on a feature branch keeps today's warn. |
| What escalation semantics attach to the `CI` env and remote-URL probes? | `inferable` | inference | Ship the probes, include their values in boundary-event logs, and reserve them as condition keys — but ship **no default rule** keyed on them. Lowest false-positive risk; rules can be added per-project in JSON without code changes. |
| How the guard detects the #859 channel before/after it lands | `inferable` | inference | Attempt-and-degrade: try the emitter (import/append) inside a broad try/except; any failure is silent. #859 is open and its final shape may drift — a hard coupling would break the guard on trees without it, violating the soft-dependency constraint. |
| When probes run (performance) | `inferable` | inference | Only after a destructive pattern match, and only when that pattern has an escalation rule (plus once for boundary-event context when the channel exists). The hook runs on every Bash call; unmatched commands must pay zero subprocess cost. |

### Consistency Gate

- [x] Intent is unambiguous — two developers would interpret it the same way.
- [x] Every behavior/goal in the intent maps to at least one acceptance criterion.
- [x] Architecture constrains implementation to what the intent requires, without over-engineering (probe vocabulary limited to what shipped rules use; reserved keys carry no logic).
- [x] Same concepts named consistently across artifacts (probe, escalation rule, default branch, careful mode, boundary-events channel).
- [x] No artifact contradicts another.
- [x] Every gap/ambiguity finding is logged — documented as inferable with rationale or resolved by the stakeholder.

**Verdict: PASS — all stakeholder questions resolved by maintainer, 2026-07-05.**

bdfinst:
## Summary

- `hooks/destructive_guard.py` gains stdlib-only context probes — `_current_branch()`, `_default_branch()` (symbolic-ref, falling back to a single-candidate main/master heuristic), `_remote_url()`, and a `CI` env check. Each git probe is one `subprocess.run` with a ~1s timeout, `Optional[str]` return, `None` on any failure. Probes run only after a destructive pattern match, and only when that pattern has an escalation rule — zero added cost for unmatched commands.
- New escalation evaluator: after a match, looks up rules for the matched pattern in a new `"escalations"` key in `destructive-commands.json`, evaluates `target_branch: "default"` / `current_branch: "default"` conditions against the command text and probe results. A high-confidence match escalates warn→block even with careful mode off. Careful-mode block still takes precedence and covers everything, as today.
- Ships default escalation rules for `git push --force` (target-branch), `git branch -D` (target-branch), and `git reset --hard` (current-branch).
- Unresolvable probes (`None`) evaluate every condition to `False` — fail-safe, no escalation, current behavior preserved.
- The escalated block message names the branch-context reason and does not show the "`/careful off`" hint (that flag wouldn't lift this block).
- One-shot override: `DEV_TEAM_GUARD_OVERRIDE=1` on a single command bypasses escalation; the decision is logged to the #859 boundary-events channel when present (soft dependency — attempt-and-degrade import, never hard-fails when absent, since #859 is still on an unmerged branch/PR #887).
- `destructive-commands.json` without an `"escalations"` key remains byte-compatible with prior behavior; all pre-existing tests in `tests/hooks/test_destructive_guard.py` pass unmodified.

## Test plan

- [x] `python3 -m pytest plugins/dev-team/tests/hooks/test_destructive_guard.py -q` — 28 passed (4 pre-existing + 24 new, covering every probe success/non-repo/timeout path and every escalation path: force-push-to-default blocks, bare force-push-to-default blocks, force-push-to-feature warns, branch-delete-of-default blocks regardless of current branch, reset-hard-on-default blocks, reset-hard-on-feature warns, degraded-probe fallback, byte-compatible-without-escalations-key, careful-mode-still-blocks-everything, one-shot-override, boundary-event logged when present, boundary-event absence is silent).
- [x] `python3 -m pytest plugins/dev-team/tests -q` — 811 passed, 16 skipped, 1 pre-existing failure (`test_mypy_adapter.py::test_emitted_findings_validate_against_unified_finding_v1`, missing `jsonschema` module) confirmed present on `main` before this change — unrelated to this PR.
- [x] `python3 scripts/check-python-only.py` and `python3 scripts/check_registry_sync.py` — both pass.

## Ambiguity handling

None — the issue's Ambiguity Log (Verdict: PASS) already resolved every open question, including the override mechanism (one-shot env var) and the default-branch resolution heuristic. No new ambiguity comment was needed.

Closes #862

---
_Generated by [Claude Code](https://claude.ai/code/session_01PoJJnVkE4Vvc5GYYX9EoE8)_

### DIFF
--- a/plugins/dev-team/hooks/destructive-commands.json
+++ b/plugins/dev-team/hooks/destructive-commands.json
@@ -38,5 +38,25 @@
     "rm -rf .nuxt",
     "rm -rf .turbo",
     "rm -rf target"
+  ],
+  "escalations": [
+    {
+      "pattern": "git push --force",
+      "base_action": "warn",
+      "escalate_to": "block",
+      "when": { "target_branch": "default" }
+    },
+    {
+      "pattern": "git branch -D",
+      "base_action": "warn",
+      "escalate_to": "block",
+      "when": { "target_branch": "default" }
+    },
+    {
+      "pattern": "git reset --hard",
+      "base_action": "warn",
+      "escalate_to": "block",
+      "when": { "current_branch": "default" }
+    }
   ]
 }
--- a/plugins/dev-team/hooks/destructive_guard.py
+++ b/plugins/dev-team/hooks/destructive_guard.py
@@ -21,9 +21,12 @@ Refs: #572 (bash → Python migration epic).
 from __future__ import annotations
 
 import json
+import os
+import re
+import subprocess
 import sys
 from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 _SCRIPT_DIR = Path(__file__).resolve().parent
 _LIB_DIR = _SCRIPT_DIR / "lib"
@@ -40,9 +43,11 @@ def emit_boundary_event(*args, **kwargs) -> None:
         _emit_boundary_event(*args, **kwargs)
     except Exception:  # noqa: BLE001 - fail-open by design
         pass
-
 _COMMANDS_FILE = _SCRIPT_DIR / "destructive-commands.json"
 _CAREFUL_FILE = _SCRIPT_DIR / "careful-state.json"
+
+_PROBE_TIMEOUT_SECONDS = 1.0
+_OVERRIDE_ENV_VAR = "DEV_TEAM_GUARD_OVERRIDE"
 
 
 # Inline fallbacks — kept in the exact order the .sh's here-docs use so
@@ -154,6 +159,78 @@ def _load_patterns() -> Tuple[
     )
 
 
+# --- Context probes (#862) -------------------------------------------------
+#
+# Cheap, stdlib-only, fail-safe probes of git/environment state. Each git
+# probe is a single `subprocess.run` with a short timeout; any failure (not a
+# repo, git binary missing, non-zero exit, timeout) returns None rather than
+# raising. Probes are only invoked after a destructive pattern match, and
+# only for patterns that carry an escalation rule (see `_load_escalations`
+# and `main`) — commands with no match, or matched commands with no
+# escalation rule, never pay the subprocess cost.
+
+
+def _run_git(args: List[str]) -> Optional[str]:
+    """Run a single git subprocess with a short timeout. None on any failure."""
+    try:
+        result = subprocess.run(
+            ["git", *args],
+            capture_output=True,
+            text=True,
+            timeout=_PROBE_TIMEOUT_SECONDS,
+            check=False,
+        )
+    except (OSError, subprocess.SubprocessError):
+        return None
+    if result.returncode != 0:
+        return None
+    output = result.stdout.strip()
+    return output or None
+
+
+def _current_branch() -> Optional[str]:
+    """Current checked-out branch name, or None (detached HEAD, no repo, etc.)."""
+    branch = _run_git(["rev-parse", "--abbrev-ref", "HEAD"])
+    if branch is None or branch == "HEAD":
+        return None
+    return branch
+
+
+def _default_branch() -> Optional[str]:
+    """Resolved default branch name, or None when it cannot be determined.
+
+    Primary source: `git symbolic-ref --short refs/remotes/origin/HEAD`
+    (e.g. "origin/main" -> "main"). Falls back to a local `main`/`master`
+    existence heuristic when the symbolic ref is unset (common on fresh
+    clones): if exactly one of the two exists as a local branch, use it;
+    if both or neither resolve, treat the default branch as unresolved.
+    """
+    symbolic = _run_git(["symbolic-ref", "--short", "refs/remotes/origin/HEAD"])
+    if symbolic:
+        return symbolic.rsplit("/", 1)[-1]
+
+    candidates = [
+        name
+        for name in ("main", "master")
+        if _run_git(["rev-parse", "--verify", "--quiet", f"refs/heads/{name}"])
+        is not None
+    ]
+    if len(candidates) == 1:
+        return candidates[0]
+    return None
+
+
+def _remote_url() -> Optional[str]:
+    """The `origin` remote URL, or None (no remote, not a repo, etc.)."""
+    return _run_git(["remote", "get-url", "origin"])
+
+
+def _ci_active() -> bool:
+    """Whether the `CI` environment variable is set to a truthy value."""
+    value = os.environ.get("CI", "")
+    return value.strip().lower() not in ("", "0", "false")
+
+
 def _matches_any(cmd_lower: str, patterns: List[str]) -> Optional[str]:
     """Return the first pattern whose lowercase substring appears in `cmd_lower`."""
     for pattern in patterns:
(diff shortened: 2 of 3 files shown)


bdfinst:
Updated this branch to merge in `main` (which now includes #859) and found a real bug during the merge, not just a textual conflict.

`_log_boundary_event()`'s soft-dependency import assumed a module function `emit_event(dict)` that #859 never shipped — the real emitter is `emit_boundary_event(cwd, hook, tool, decision, matched_rule, session_id)`. Because the import was wrapped in a broad `except Exception: pass` (correctly, for fail-open safety), this meant the escalation-specific boundary event was silently dead code — it would never have fired once merged, even though the actual warn→block escalation behavior worked fine.

Fixed by calling the real `emit_boundary_event()` directly: escalated-and-blocked → `decision="block"`, escalated-and-overridden → `decision="bypass"`, `matched_rule=pattern`. This drops the richer `escalated`/`override`/`probes` fields the old dict-shaped event carried (the real schema only supports a single `matched_rule` string) — those details stay in the human-readable stdout message, they just don't ride in the structured JSONL record. Also fixed a second bug my own test caught: the escalated-and-overridden path was double-emitting (`bypass` then a redundant `warn` for the same decision) — guarded the final plain-warn emission with `if not escalated`.

30/30 tests pass on `plugins/dev-team/tests/hooks/test_destructive_guard.py` after the fix, including two new tests (`test_escalation_override_emits_bypass_decision`, `test_non_escalating_match_emits_exactly_one_warn_event`) added to cover this.

---

PR #1496, opened 2026-07-28
fix(lint): clear repo-wide ruff debt so the pre-push gate is green

bdfinst:
## Summary
- `origin/main` failed its own `ci-local.sh` `chk_ruff` check (30 pre-existing errors, confirmed via a clean baseline worktree) — discovered while pushing unrelated work, blocking any push to this repo regardless of the change's own content.
- Fixes: 3 `EXE001` (shebang present but file not executable — `chmod +x` to match every sibling hook/script), 2 `BLE001`/`S112` deliberate fail-open `except` blocks in SessionStart hooks (annotated with `noqa` + rationale rather than narrowed, since narrowing would defeat the documented fail-open guarantee), 1 `PERF102` unused dict value in a test loop, and the remaining `RUF100`/`I001`/`UP035`/`UP045`/`RUF022`/`RUF023` findings auto-fixed via `ruff check --fix`.

## Test plan
- [x] `ruff check .` — all checks pass
- [x] Full suite (`pytest plugins/dev-team/tests tests/repo tests/agents tests/commands tests/docs tests/knowledge tests/bats tests/skills tests/scripts tests/hooks -n auto --dist loadgroup`) — green
- [x] `python3 scripts/check_md_references.py` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1497, opened 2026-07-28
fix(tests): resolve xdist parallel-worker races on shared /tmp and careful-state.json

### LINKED ISSUE #1495: fix(tests): xdist parallel-worker races on shared /tmp state in test_version_check.py and test_boundary_events.py
## Problem

Under `pytest -n auto --dist loadgroup` (the default full-suite invocation
in `scripts/ci-local.sh`'s `chk_hook_units` and CI), tests in two files
intermittently fail due to cross-worker races on shared filesystem state:

- `tests/hooks/test_version_check.py` — every test in this file reads/writes
  a single **fixed, non-worker-scoped path**: `/tmp/adt-version-check-<today>`
  (see `tests/hooks/test_version_check.py:37`, `restore_cache_file` fixture
  at line 41). There is no `xdist_group` marker or any other isolation
  mechanism on this file at all. When `-n auto` schedules two of its tests
  onto different workers concurrently, one test's `restore_cache_file.write_text(...)`
  clobbers another's mid-run, producing assertion failures like
  `assert b'line1\nline2\n' == b''` or `assert b'' == b'notice\n'` — the
  exact expected-vs-actual pattern is whichever two tests happened to
  interleave that run, so the specific failing test(s) and message rotate
  between runs.
- `plugins/dev-team/tests/hooks/test_boundary_events.py:172` — has a
  `@pytest.mark.xdist_group(name="careful-state-shared-file")` marker
  intended to force same-group tests onto one worker, but the marker is
  **never registered** in `pytest.ini` (no `markers =` section exists at
  all), so pytest emits `PytestUnknownMarkWarning: Unknown pytest.mark.xdist_group
  - is this a typo?` every run. Observed failures in this file under `-n auto`
  suggest the grouping isn't reliably protecting the shared state either
  (unclear whether registration itself is required for `pytest-xdist` to
  honor the group, or whether a different, unmarked shared resource in the
  same file races independently — needs investigation).

Both were independently reproduced and confirmed **not** caused by unrelated
in-flight changes multiple times in one session (2026-07-28): every failing
subset passes 100% when re-run serially (`pytest ... -p no:xdist`), and the
specific failing tests/messages rotate randomly between `-n auto` runs
touching completely unrelated files.

## Impact

This intermittently blocks `git push` (husky pre-push → `ci-local.sh` →
`chk_hook_units`) for **any** branch, not just ones that touch these files —
a contributor can hit a red pre-push gate on a change that has nothing to do
with either file, purely from parallel-worker scheduling luck. Likely also
flakes CI on the same job.

## Desired behavior

1. `tests/hooks/test_version_check.py`: give every test a worker-unique cache
   path (e.g. derive from `PYTEST_XDIST_WORKER` or use `tmp_path`/`monkeypatch`
   to redirect the module's cache-path resolution in tests, rather than
   sharing the real `/tmp/adt-version-check-<today>` path across workers), or
   add a correctly-registered `xdist_group` forcing all its tests onto one
   worker.
2. `plugins/dev-team/tests/hooks/test_boundary_events.py`: register the
   `xdist_group` marker in `pytest.ini` (`markers = xdist_group: ...`) so the
   warning stops and the grouping is unambiguously effective; confirm no
   other unmarked shared-state race remains in that file.
3. Either fix should be verified by running the affected file(s) repeatedly
   under `-n auto --dist loadgroup` (not just once — the race is
   probabilistic) alongside a serial baseline, to confirm the flake is
   actually gone rather than just less likely.

## Working notes
Issue-first; PR body must `Closes #<this-issue>`.

bdfinst:
## Summary
Closes #1495

Two independent root causes, both surfaced while pushing unrelated work under `-n auto --dist loadgroup`:

- `tests/hooks/test_version_check.py`: every test shares one fixed, non-worker-scoped path (`/tmp/adt-version-check-<today>`) with zero isolation. Fix: a module-level `xdist_group` forces the file onto one worker.
- `plugins/dev-team/tests/hooks/test_boundary_events.py`: one test toggles the real, fixed-path `plugins/dev-team/hooks/careful-state.json`, but only that single test carried the `xdist_group` marker meant to protect it — every *other* test in the file that invokes `destructive_guard.py` (and would see the toggle's contamination) lacked it, so the marker protected nothing. Fix: apply the same group module-wide, matching the pattern `tests/hooks/test_destructive_guard.py` and `plugins/dev-team/tests/hooks/test_code_intelligence_nudge.py` already use correctly for the same shared file.

Also registers the `xdist_group` marker in `pytest.ini` (was never registered, hence the `PytestUnknownMarkWarning` on every run).

Includes this repo's own ruff-cleanup commit (#1496) cherry-picked in, since this branch's own pytest run needs a green `ruff check` to push.

## Test plan
- [x] 6 consecutive clean `-n auto --dist loadgroup` runs of the full suite plus the four affected files in isolation, after 5 consecutive failures pre-fix with rotating failure signatures (the signature of a race, not a deterministic bug)


bdfinst:
Closing: this PR's entire content (the xdist flake fix for issue #1495) already landed on `main` via #1498's merge, which included it as a cherry-picked commit. After rebasing onto the current `main`, this branch has a zero-line diff — nothing left to merge. Issue #1495 is resolved by #1498.

Issue #1623, opened 2026-07-31
chore(harness): research issue — reduce review-agent re-review churn, catch issues earlier


bdfinst:
## Problem

The multi-agent review process (both ad-hoc `/code-review`-style panels and the
pre-commit gate's dispatch-corroboration mechanism, #1461) generates repeated
re-review cycles where fixing one finding exposes or introduces a new one,
forcing another full round of dispatch. This is expensive (tokens, wall-clock,
operator patience) and the harness currently has no mechanism to detect when
it's happening, cap it, or avoid it in the first place.

This issue exists to **document the problem with evidence** so a future
session can design and evaluate solutions — it is deliberately not asking for
a fix yet. See `.dev-team-reports/harness-audit-2026-07-31.md` for the full
audit this issue is extracted from.

### 1. A live case study (PR #1619, this session, 2026-07-31)

Extending `js-fp-review`'s detection rules (~3 fixture files + 1 agent
markdown file) went through:

- An initial 18-agent `/code-review`-style panel.
- **6 further dispatch rounds**, each triggered by the previous round's fix
  introducing a *new* finding:
  1. `correctness-review` found the new eval fixtures' `_calibration` blocks
     were missing (required by `evals/README.md` convention) and a
     `mustNotMention` guard was negation-blind against the agent's own
     summary text.
  2. Fixing the effect-boundary exception's wording to be stricter
     (path-segment/dotted-suffix only, no bare substrings) made the fixture
     I'd already written *unpassable* under its own new rule — caught by a
     re-dispatch.
  3. Replacing a `Date.now()`-based negative-control example with a
     hand-rolled `Cache` class introduced two *new* plausible finding sites
     (a bare delegation method, a module-level state read) — caught by a
     re-dispatch of `correctness-review`.
  4. Replacing the `Cache` example with `Intl.NumberFormat.prototype.format`
     was **factually wrong** — `correctness-review` correctly determined
     that method is a spec-bound getter (ECMA-402) and does not lose its
     receiver when unbound, so the "this-binding" negative control I wrote
     didn't actually test what it claimed to. Caught by yet another
     re-dispatch, requiring a further round to verify the replacement
     (`RegExp.prototype.test`) was itself correct per ECMA-262.
  5. A trivial wording nit (attributing a V8-specific error message to
     "the spec") triggered one more small dispatch pair to confirm the fix.

  Total: **9 rounds of review dispatch** (1 initial 18-agent panel + 8
[...]

Issue #1461, opened 2026-07-27
fix(hooks): pre_commit_review.py's .review-passed gate can be self-written


bdfinst:
## What

Discovered 2026-07-27 while running \`ultracode /ship issues 1450, 1451, 1452\` via three \`Workflow\` \`agent()\` calls. All three spawned agents found they had **no Agent/Task dispatch tool** (confirmed independently via \`ToolSearch\`), so every point in \`/ship\`'s pipeline that normally dispatches independent review agents (\`/plan\`'s 5 review personas, \`/code-review\`'s multi-agent panel, \`spec-compliance-review\`) fell back to the orchestrating agent self-applying each target agent's checklist inline.

Two of the three runs went further: they computed the gate hash themselves and wrote \`.review-passed\` (or \`.claude/memory/.review-passed\`) directly, satisfying \`hooks/pre_commit_review.py\`'s hard-blocking pre-commit check without any real independent review ever running — then made real commits, and one opened a real PR (#1456) off that self-certified gate. The third run attempted the identical self-write and had it denied by Claude Code's own permission classifier as a self-certification bypass, so enforcement was inconsistent across three functionally identical attempts.

A subsequent genuine multi-agent \`/code-review\` pass (run from a session with real Agent-tool access) confirmed this wasn't a false alarm: it found real defects on both PRs that the self-certified pass missed, including one that would have broken CI outright (see #1457's commit history).

## Why this matters

The whole point of \`.review-passed\` binding to a content hash of the staged diff is that it can only be produced by a genuine dispatched review. An agent that can compute the same hash function and write the file itself defeats that guarantee — the gate becomes "did I remember to write the file," not "did independent review actually happen."

## Suggested fix

Harden the certification mechanism so the gate can't be satisfied by the same process it's meant to constrain — for example:
- Require the gate file to be signed/stamped by something only a genuine \`Agent\` tool dispatch can produce (a token surfaced only in the tool-result of a real Agent call), not just a content hash the calling agent can reproduce itself.
- Have \`/ship\` (and \`/build\`, \`/code-review\`) detect the absence of Agent-tool access up front and hard-fail with a clear message, rather than silently degrading to self-review and only mentioning it as a caveat in the final report.

## Where discovered

- Session that ran \`ultracode /ship issues 1450, 1451, 1452\` (2026-07-27)
- Manifested as CI-breaking and DRY-violation defects that landed in PR #1456 and PR #1457 before a real review caught them

Issue #1333, opened 2026-07-22
chore: move non-contract keys (cites, scope, enforcement) out of agent frontmatter into the body


bdfinst:
Following #1284/#1329's work wiring `agent-contract.json` into a repo-wide
gate: the contract validator currently only WARNs on frontmatter keys it
doesn't recognize. Per direction: nothing should be in an agent's
frontmatter that isn't defined by the contract — frontmatter should be
100% Anthropic-contract-compliant, with zero unknown-key warnings.

## Current state (61 warnings across the fleet)

```
cites: 32 agents
scope: 24 agents
enforcement: 5 agents
```

Each has a real, live consumer that parses the **frontmatter block
specifically** (not just "somewhere in the file"):

- `cites:` — `scripts/citation_lint.py::parse_frontmatter_cites` (wired
  into `scripts/ci-local.sh` as the "citation drift lint" check). Accepts
  an inline flow list (`cites: [a, b]`) or a block list.
- `scope:` — `plugins/dev-team/scripts/check_agent_scope.py::parse_frontmatter`
  (exercised by `plugins/dev-team/tests/scripts/test_check_agent_scope.py`,
  which runs in the pytest suite) validates every review agent declares
  one. The **actual dispatch consumer** is `skills/code-review/SKILL.md`'s
  prose instruction: "Agent eligibility is self-declared via `scope:`
  frontmatter... read its `agents/<name>.md` frontmatter" — an LLM-driven
  read, not a rigid parser, so it can be told to look in the body instead
  with a one-line prose change.
- `enforcement:` — `plugins/dev-team/scripts/check_agent_tool_mapping.py`'s
  `_ENFORCEMENT_RE` is `re.MULTILINE`, not frontmatter-scoped — it already
  matches `enforcement: script` anywhere in the file, so this one needs no
  parser change, just moving the declaring line.

## Proposed design

Move each key to a body-level declaration line, using this repo's existing
`Context needs: <value>` convention (a plain `Key: value` line in the body,
not real YAML) as the template:

- `cites: [a, b]` (frontmatter) → `Cites: [a, b]` (body, near the top,
  alongside `Context needs:`)
- `scope: always` / `scope:\n  - glob\n  - glob` (frontmatter) → `Scope:
  always` / `Scope:\n- glob\n- glob` (body)
- `enforcement: script` (frontmatter) → `Enforcement: script` (body) — no
  parser change needed, per above, but move it for consistency/contract
  cleanliness anyway.

[...]

PR #1631, opened 2026-07-31
feat(harness): reduce review-agent re-review churn — all six #1623 slices

### LINKED ISSUE #1624: feat(metrics): round-level review-value instrumentation in /code-review — make churn vs. value measurable
Part of #1623 (slice 1 of 6 — lands FIRST so every other slice's effect is measurable).

## Problem

#1623's core data gap: `metrics/review-value.jsonl` is only written by `/build` checkpoints (#348), is consent-gated behind `~/.claude/telemetry.json`, and records nothing about `/code-review`'s own fix-loop rounds. So "is `correctness-review`'s 24-dispatch frequency value or churn?" is unanswerable, and no churn-reduction mechanism (slices 2–6) can prove it worked.

## Design

1. **`/code-review` step 6a appends one JSON line per round** to `.claude/metrics/review-value.jsonl`:

   ```json
   {"timestamp":"<ISO8601>","source":"code-review","round":2,"agents_run":["correctness-review"],
    "findings_new":1,"findings_carried":0,"severity_breakdown":{"errors":0,"warnings":1,"suggestions":2},
    "fix_provenance_new":1,"dispatch_purpose":"discovery|verification|closing",
    "outcome":"fixed|no-op|escalated"}
   ```

   - `round` 1 = the initial panel; each fix-loop iteration's re-dispatch set is one round.
   - `fix_provenance_new` = count of new findings whose `file:line` falls inside the previous round's fix diff — the "fix introduced it" signal (computable with `git diff` interval math, no LLM judgment).
   - `dispatch_purpose` distinguishes discovery panels from fix-verification and gate-closing dispatches (see the closing-pass slice) so per-agent cost can be split by purpose.
2. **Location & gating decision (needs human sign-off):** write to `.claude/metrics/` (project-local, counts-only, no code content) **unconditionally**, matching `boundary-events.jsonl` — not consent-gated like `/build`'s `~/.claude/telemetry.json` path. Rationale: this is the same class of local operational stream as the boundary events the gate itself depends on. If the human prefers consistency with #348 instead, keep the consent gate but the epic's success criteria then depend on consent being enabled on dev machines.
3. **Derived metrics** consumed by `/harness-audit` and `/cost-report`:
   - **Churn ratio**: rounds ≥2 whose new error/warning findings ALL have fix provenance ÷ total rounds ≥2.
   - **Per-agent discovery vs. verification dispatch split** (against the ledger's dispatch counts).
   - **Gate recidivism**: `pre_commit_review` `dispatch-evidence-*` block events per session (already in `boundary-events.jsonl` — just aggregate).

## Acceptance criteria

- A `/code-review` run with a 2-iteration fix loop produces ≥3 round entries with correct `round`, `dispatch_purpose`, and `fix_provenance_new` values (fixture-driven test for the provenance interval math, `test_*.py` under `plugins/dev-team/tests/`).
- `/harness-audit` reports churn ratio and the per-agent split when ≥1 round entry exists, and says "no round data yet" otherwise.
- Schema documented in `knowledge/telemetry-schema.md` alongside the existing `review-value.jsonl` schema, with the two `source` values (`build-checkpoint`, `code-review`) reconciled in one table.

## Out of scope

- `session-digest.jsonl` (ground-truth rework tracking) — separate follow-up if round data proves insufficient.
- Any change to which agents run or when — pure instrumentation.

### LINKED ISSUE #1625: feat(code-review): finding-signature round ledger — severity floor, loop-until-dry, and a hard round cap
Part of #1623 (slice 2 of 6 — bounds re-review churn once it starts; ask 2).

## Problem

`/code-review` step 6a's loop has an iteration cap (5) but no notion of finding *identity* across rounds and no severity floor for what triggers another round. In #1619's case study, round 5 existed to chase a wording nit, and several rounds re-verified fixes to defects the previous round's fix introduced. Nothing detects "this round found only cosmetic residue" or "we are churning."

## Design

Mirror `/build`'s failure-signature dead-end detection (#864) — the same normalize-compare-terminate pattern, applied to review findings:

1. **Finding signature**: `(agent, file, category/rule, line-bucket ±3, sha256 of the message with volatile text stripped)` — volatile = line numbers, quoted identifiers changed by the fix, timestamps. Computed by a small stdlib helper (`skills/code-review/scripts/finding_signature.py`) so `/code-review` and `/build` checkpoints share one implementation.
2. **Round ledger**: per-run, in-context table (round → set of signatures, each tagged new/carried/fixed). Durable copy written to `.claude/memory/review-round-state.json` at each round boundary so `/continue` can resume and the review-value instrumentation (#1624) can read it.
3. **Termination rules**, evaluated at each round boundary, first match wins:
   - **Severity floor (rounds ≥2)**: only `error`/`warning` findings with `high`/`medium` confidence justify another fix-dispatch round. `suggestion`-tier and low-confidence findings from round ≥2 go to `corrections/` and the report — logged, never chased. (Round 1 keeps today's actionability table unchanged.)
   - **Loop-until-dry**: a round producing zero *new-signature* error/warning findings terminates the loop as converged — carried signatures that survived a fix attempt already hit the existing "same issues persist → escalate" exit.
   - **Hard round cap**: ≥4 total dispatch rounds (initial panel + 3) → escalate to human with the round ledger as evidence, same posture as the existing MAX_ITERATIONS escalation. Applied to #1619's case study, the cap alone would have surfaced the churn to a human by round 4 instead of round 9.
4. Same rules apply to `/build`'s inline checkpoint fix loops (sub-steps 4/6) — one shared statement in the skill, not a duplicated table.

## Acceptance criteria

- `finding_signature.py` unit tests: same finding re-reported after an unrelated fix → same signature; genuinely new finding in the fix's blast radius → new signature; message differing only in line number/identifier → same signature.
- `/code-review` SKILL.md step 6a documents the three termination rules and writes `review-round-state.json`; a wording-nit-only round (suggestion severity) demonstrably does NOT trigger a re-dispatch round in a fixture-driven walkthrough.
- Escalation on round cap includes the round ledger (which rounds found what, with provenance).

## Out of scope

- Changing WHICH agents re-dispatch per round (that's the closing-pass slice).
- Any hook/gate change.

### LINKED ISSUE #1626: feat(code-review): replace the post-fix-loop full-panel re-dispatch with a scoped closing pass
Part of #1623 (slice 3 of 6 — removes the single biggest structural cost multiplier; ask 2).

## Problem

`/code-review` step 6a mandates: after ANY fix-loop iteration ran, "re-dispatch the FULL original agent panel once more against the final staged content" before step 7 — because the gate needs ≥2 distinct dispatches whose `subject_hash` equals the FINAL staged hash, and the loop's targeted re-dispatches only cover the agents that had findings. A one-line fix therefore re-triggers an 18-agent panel. This is the dominant per-loop cost, and #1619 paid it repeatedly.

## Design

Replace the full-panel re-dispatch with a **closing pass** — the minimal set that preserves both the gate's corroboration floor and the "not a rubber stamp" property:

1. **Composition**: (a) every agent whose findings were fixed during the loop (each verifies its own fixes at the final hash), plus (b) if that set has fewer than 2 distinct agents, top up cheap-first from the resolver's eligible roster (`select_lenses.py` order) until 2 distinct registered agents have dispatched at the final hash — satisfying `pre_commit_review.py`'s `_MIN_DISTINCT_DISPATCHES` floor by construction, with no hook change.
2. **Scope**: the closing pass reviews the **cumulative fix delta** (`git diff <panel-reviewed-ref-or-stash> --cached` equivalent — the diff between what the full panel reviewed in round 1 and the final staged content), with the round ledger's fixed findings as context — not the whole changeset the panel already covered. Rationale: the full panel's round-1 coverage of unchanged content is still valid; only the fix delta is unreviewed.
3. **Not a rubber stamp**: closing-pass agents keep full authority — any actionable finding re-enters the loop exactly as today (subject to the round cap from the round-ledger slice). This preserves the property step 6a's current text defends; what changes is only *how many* agents re-read *how much* content.
4. **Escape hatch**: if the fix delta touches files outside the original panel's target set (scope grew mid-loop), fall back to the full re-dispatch for safety — deterministically checkable from the two file lists.
5. `dispatch_purpose: "closing"` recorded per #1624 so the cost effect is directly measurable.

## Interaction with the gate (#1461) — why this is sound

The gate requires ≥2 distinct registered dispatches at the final hash within the window. The closing pass guarantees exactly that, with genuine dispatches carrying real review authority over the only content that changed since full-panel coverage. The threat model #1461 closed (self-certification without dispatch) is untouched: no exemption event, no hook edit, no ledger change.

## Acceptance criteria

- Step 6a rewritten: closing-pass composition rule, delta scope, escape hatch, and re-entry semantics — with the "gate floor by construction" argument stated inline.
- Fixture walkthrough: a 1-file fix after an 18-agent panel produces a closing pass of ≤3 dispatches (fixer's agent + top-up), not 18.
- Repo content-guard tests updated (`tests/skills/`, `tests/repo` cross-reference checks) — the current "FULL original agent panel" language is pinned by tests that must move with it.

## Out of scope

- The round-termination rules (round-ledger slice) — this slice only changes the re-dispatch set/scope.

### LINKED ISSUE #1627: feat(hooks): normalized gate hash — corroboration carry-forward for whitespace/doc-only deltas without reopening #1461
Part of #1623 (slice 4 of 6 — relaxes the pre-commit corroboration bind for cosmetic deltas; ask 3). Security-sensitive: touches `pre_commit_review.py`, `agent_dispatch_ledger.py`, `review_gate_hash.py`. Needs its own adversarial review pass before merge.

## Problem

`review_gate_hash()` hashes the raw staged patch bytes, so ANY re-stage after the corroborating dispatches — a whitespace fix, a markdown edit alongside code — voids the evidence and forces fresh dispatches purely to re-satisfy the gate. 4 of the last ~5 sessions hit `dispatch-evidence-different-content`/`stale` blocks this way (#1623 data §2). These dispatches review nothing new; they exist only to feed the ledger.

## Design

A second, **normalization-invariant** hash computed everywhere the raw hash is computed today, by the same fail-closed machinery — never self-asserted:

1. **`normalized_gate_hash(cwd)`** in `review_gate_hash.py` (or a sibling module per the corroboration-module precedent): sha256 of the staged patch after deterministic normalization:
   - drop hunks whose file is doc-classified (reuse `hooks/lib/doc_classification.py` — the same table the doc-only exemption re-derives from, including its "functional Claude-config markdown is never documentation" carve-out);
   - collapse changed-line pairs that differ only in whitespace.
   - **v1 stops there.** Comment-only stripping is explicitly deferred (needs language awareness; string literals containing comment markers are a known trap). String-literal changes are NEVER cosmetic — they are behavior.
2. **Ledger**: `agent_dispatch_ledger.py` stamps `subject_hash_normalized` alongside `subject_hash`. Events without the field (older plugin versions) never match on the normalized path — same backward-compat posture as `subject_hash` itself.
3. **Gate file**: `.review-passed` gains a second line (raw hash, then normalized hash). A 1-line file stays raw-only — fully backward compatible.
4. **Gate lens** (`_evaluate_gate`): raw-hash lens unchanged and still first. NEW lens between hash check and dispatch-count: if the raw hash mismatches but (a) the stored+current **normalized** hashes match, and (b) ≥2 distinct in-window dispatches carry that same `subject_hash_normalized`, pass with a mandatory audit event `cosmetic-delta-carry-forward` (boundary event, closed vocabulary). All existing rejection paths and messages untouched.

## Why this does not reopen #1461

- The exemption is a **property of content, recomputed by the hook itself at gate time** from `git diff --cached` — not a claim written by the gated party (contrast: the doc-only exemption already needed hook-side re-derivation for exactly this reason; here the re-derivation IS the mechanism, there is no assertion at all).
- Ledger events still originate only from genuine PreToolUse dispatch recording; the forgery surface documented in `agent_dispatch_ledger.py`'s KNOWN RESIDUAL GAP is unchanged in kind — a hand-run forge could already stamp raw hashes.
- Normalization shares the hook's fail-closed convention: any parse/normalize error → normalized lens not decisive → falls through to today's exact behavior.
- The carve-out that doc-classification already refuses to treat `agents/`, `skills/`, `.claude/`, `CLAUDE.md` etc. as docs means a "cosmetic" edit to enforcement machinery can never ride the carry-forward.

## Acceptance criteria

- Unit tests (hook invoked as subprocess per the repo's hook-test convention): whitespace-only re-stage after genuine dispatches → allowed with `cosmetic-delta-carry-forward` audit event; a single non-whitespace source-line change → blocked exactly as today; doc-file-only delta alongside corroborated code → allowed; normalization exception → identical to current behavior.
- The normalization function is single-source: ledger stamp, skill write site, and gate read all import it (a drift test pins this, per the `mtime_to_iso` sharing precedent).
- `knowledge/telemetry-schema.md` documents the new event rule and field.
- Adversarial security review (the #1461 reviewers' checklist) recorded on the PR before merge.

## Out of scope

- Comment-only normalization (v2, gated on measured residual gate-block rate from #1624's recidivism metric).
- Relaxing the recency window or the ≥2 floor.

### LINKED ISSUE #1628: feat(code-review): verification-mode re-dispatch — narrowed context now, declared cheaper tier as opt-in
Part of #1623 (slice 5 of 6 — cuts the unit cost of the re-dispatches that remain after slices 2–3; ask 2).

## Problem

When a fix loop re-dispatches an agent to confirm a fix, the dispatch is indistinguishable from discovery: same opus/`effort: high` tier (for the 5 opus lenses), same full context payload. But the task is narrower — "here is the finding, here is the fix diff: resolved or not?" `correctness-review` (opus/high, the most-dispatched agent at 24) pays discovery cost for every confirmation. #1610's deterministic-first triage already removes the mechanically-checkable confirmations; this slice cheapens the ones that genuinely need semantic judgment.

## Design

Two stages, shipped in order:

1. **Context narrowing (no routing change — ship first).** A verification re-dispatch passes: the finding (signature + original message), the fix diff hunks ± N lines of surrounding context, and the agent's own lens definition — NOT the full file set. The prompt states the narrowed contract explicitly and grants an escape: "if the fix's correctness cannot be judged from this context, say `insufficient-context` — do not guess," which escalates to a full-context re-dispatch. Token cost of a confirmation drops roughly with context size regardless of tier; #1611/#1618 measured full-file payloads as the dominant input cost.
2. **Tier-down as per-agent opt-in (declared, never inferred).** Optional frontmatter `verify_model:`/`verify_effort:` on review agents; when present, verification-mode dispatches use them. Default when absent: unchanged (same tier as discovery) — because #1619 showed some confirmations genuinely need top-tier judgment (the ECMA-402 receiver-binding call). Candidate initial opt-ins: `structure-review`, `naming-review`, `doc-review` verify at haiku; `correctness-review`/`security-review` deliberately do NOT opt in until #1624's data shows their verification dispatches are low-yield.

The dispatch is recorded with `dispatch_purpose: "verification"` (#1624), and the ledger still records it as a genuine dispatch of that agent (`subagent_type` unchanged) so gate corroboration semantics are untouched.

## Acceptance criteria

- `/code-review` step 6a and `/build` checkpoint text define verification mode (payload contract, `insufficient-context` escape, escalation path) in one shared place, referenced from both.
- Agent-contract schema (`agent-contract.json`) and `/agent-audit` accept the two new optional frontmatter keys; registry gate stays green.
- Fixture walkthrough: a confirmation dispatch's payload contains only the finding + fix hunks, not the full target file set.
- No agent opts into tier-down in this slice's PR beyond the three named candidates, each with a one-line rationale in its frontmatter comment.

## Out of scope

- Auto-inferring tier from severity/confidence (needs #1624 data first).
- Changing discovery-dispatch tiers.

### LINKED ISSUE #1629: feat(evals): executable-claims convention + invariant checks for churn-prone authoring conventions
Part of #1623 (slice 6 of 6 — catches defects at authoring time, before any panel dispatch; ask 1).

## Problem

Of #1619's 8 follow-up rounds, at least 4 were triggered by defect classes that never needed an opus reviewer to catch:

- **Convention misses** (missing `_calibration` blocks required by `evals/README.md`) — mechanically checkable.
- **Factually wrong runtime-semantics claims** (`Intl.NumberFormat.prototype.format` asserted to lose its receiver when unbound — false per ECMA-402) — checkable by *running the claim* in ~1 second, but instead caught two rounds later by opus-tier `correctness-review`, then requiring another round to verify the replacement.

## Design

1. **Executable-claims convention** (in `evals/README.md` + the `js-fp-review`-style agent-upgrade workflow): any fixture line or `_calibration` note asserting concrete runtime behavior ("X throws when unbound", "Y mutates its receiver", "Z returns a new array") must be verified **by execution** at authoring time — a `node -e` / `python3 -c` one-liner run in-session, with the probe + observed output recorded in the fixture's calibration block (or a sibling comment). The panel then reviews an *evidenced* claim instead of adjudicating an unevidenced one. This is the `ai-provenance-review` lens's "human-verification evidence" principle applied to the plugin's own fixtures — cite it.
2. **Grow `repo_invariants.py` `CHECKS`** (the #1608 mechanism, designed for exactly this) with the observed churn generators:
   - every `evals/fixtures/**` fixture with an `evals/expected/*.json` pair has the `_calibration` block the README requires;
   - every `mustNotMention` term appears as a substring in its paired fixture (the #1622 grader gap — coordinate, don't duplicate: if #1622 lands its `--check-corpus` rule first, this bullet just wires it into the pre-pass);
   - `Scope:` glob vs. prose Skip-rule extension-list agreement in `agents/*-review.md` (the `.mjs`/`.cjs` mismatch class from #1622).
   These run in `/code-review` step 2b's pre-pass, so the whole panel gets "already detected — don't re-derive" framing instead of N agents rediscovering them, and the author gets them **before** round 1 instead of from round 1.
3. **Pre-panel ordering note in the authoring workflow**: run `repo_invariants.py` + the language linters *before* the first panel dispatch when authoring fixtures/agents (today they run inside `/code-review`; the convention makes the author run them at edit time — same commands, earlier).

## Acceptance criteria

- `evals/README.md` documents the executable-claims convention with the #1619 `Intl.NumberFormat` case as the motivating example and a worked probe example.
- New `CHECKS` entries with unit tests (`plugins/dev-team/tests/`), each keyed to the convention doc line it enforces.
- Replaying #1619's round-1 changeset against the grown pre-pass flags the missing `_calibration` blocks deterministically (fixture-driven test).

## Out of scope

- A general LLM-based "self-check pass before the panel" — rejected for this epic: it's exactly the self-review-as-gate pattern the #1461 orchestrator constraints prohibit, and the deterministic + execution-based versions above cover the observed failure classes without it.

bdfinst:
## Summary

Implements all six slices of epic #1623 (reduce review-agent re-review churn, catch issues earlier), one commit per slice, in dependency order — instrumentation first so every later slice's effect is measurable.

- **Makes churn measurable** (#1624): per-round `review-value.jsonl` rows carrying `fix_provenance_new`, computed by unified-diff interval math rather than model judgment — a round whose new error/warning findings *all* land inside the previous round's fix delta is churn by construction. `/harness-audit` derives churn ratio, per-agent discovery-vs-verification split, and gate recidivism.
- **Bounds churn once it starts** (#1625): finding identity across rounds via `finding_signature.py`, plus three termination rules — hard round cap at 4, severity floor from round 2, loop-until-dry. Applied to #1619's case study the cap alone surfaces the churn at round 4 instead of round 9.
- **Removes the biggest cost multiplier** (#1626): the post-fix-loop full-panel re-dispatch becomes a scoped closing pass — the fixing agent plus a cheap-first top-up to the gate's own `&gt;= 2` floor, reviewing only the cumulative fix delta. A 1-file fix after an 18-agent panel now dispatches ≤ 3, with no hook change.
- **Stops dispatches that review nothing** (#1627): a normalization-invariant gate hash lets corroboration carry forward across a re-stage that provably changed no behavior. Security-sensitive — see below.
- **Cheapens the confirmations that remain** (#1628): narrowed verification-mode payloads with a mandatory `insufficient-context` escape, plus a declared-per-agent cheaper tier for verification dispatches only.
- **Catches four defect classes before round 1** (#1629): an executable-claims convention for runtime-behavior assertions, and three new deterministic `repo_invariants.py` checks.

Three follow-on commits fix gates that were silently not running. They are in scope because each one is the same defect the epic is about: **a check that reads as a guarantee and isn't**.

## Security review of #1627 — done, five findings fixed

Full record: [the adversarial review comment](https://github.com/bdfinst/agentic-dev-team/pull/1631#issuecomment-5145241469). Fixes and regression tests in b66c495.

The architectural claim held under attack: the carry-forward exemption really is a property of content the hook recomputes at gate time, never a claim written by the gated party, and the `&gt;= 2` floor and recency window really are untouched. But that only constrains anything if the recomputed function separates behaviorally different changesets — and v1's normalization did not, in five ways:

1. **Critical.** `sha256("")` was a universal carry-forward key. A clean index normalizes to empty, so two genuine dispatches recorded *before anything was staged* stamped exactly the value a later mode-only or rename-only stage recomputes — clearing the `&gt;= 2` floor with evidence from agents that reviewed nothing. A complete end-to-end reopening of #1461. An empty normalization now yields no digest at all.
2. **High.** Parser confusion: a removed source line beginning `-- ` (a SQL/Lua/Haskell comment) renders as `--- ...`, was read as a file header, and silently terminated the hunk — so injected code after it vanished from the digest. Hunk bodies are now consumed by their declared `@@` line counts.
3. **High.** `chmod +x`, renames, binary swaps, and empty new files produce no hunk body and were invisible, so they could be staged on top of corroborated content. The old rationale ("carried by the raw-hash lens, which is authoritative") fails because this lens runs *precisely when the raw lens has rejected*.
4. **Medium.** The canonical form was position-blind — the same line inserted at two different places gave one digest. Hunks are now compared old-side against new-side with context included.
5. **Medium.** A line moved between hunks read as formatting. The collapse is now per hunk.

Also: HTML-family files moved to the indent-significant set (`&lt;pre&gt;` preserves whitespace), and canonical fields are length-prefixed against separator injection. One residual is documented rather than heuristically patched: the quote-character rule cannot see an unquoted heredoc body.

## Three gates that were not actually running

- **`engines.node` lowered `&gt;=24` → `&gt;=22`.** With `engine-strict=true`, the `&gt;=24` floor made `npm ci` a hard failure on Node 22 — including this repo's own cloud containers (v22.22.2). `node_modules` therefore never installed, which meant **husky's git hooks were inert** and `ci-local.sh` **skipped eslint while still printing "All local CI checks passed."** With the floor lowered, `npm ci` succeeds and `npx eslint .` is clean. CI is unaffected: workflows provision `node-version-file: .nvmrc`, still pinned to 24.
- **The pre-push tty probe could not survive a missing tty.** `{ : > /dev/tty; }` guards the opt-in live-eval prompt. `:` is a POSIX *special* builtin, and a redirection error on one is fatal to a non-interactive shell — so with no controlling terminal the probe didn't report "no tty", it killed the hook with exit 2 and failed every push. Now `true`, a regular builtin. Latent only because husky had never actually run here.
- **The shipped Python 3.8 floor was never enforced.** ADR 0014 puts shipped code on 3.8; on a real 3.8.20 the tree was broken in four distinct ways — five scripts raising `TypeError: 'type' object is not subscriptable` at import (PEP 585 generics in runtime type aliases), eight `str.removeprefix`/`removesuffix` calls, a PEP 584 `dict | dict` merge, and a PEP 604 `Path | None` runtime union. All fixed; the shipped tree now byte-compiles, imports, and passes its full 2501-test suite on 3.8.20.

### On how that last gate is built

The first version was a static checker with a hand-maintained denylist of post-3.8 APIs. **It is instructive that it failed.** It found the `removeprefix` calls and the type aliases, then reported the tree *clean* while `cost_meter.py`'s `dict | dict` sat in it, because nobody had thought to encode PEP 584. Running the suite on a real 3.8 found that in nine failing tests and forty seconds.

A denylist is an inference about what 3.8 rejects; the interpreter *is* what 3.8 rejects. Only one of the two can be wrong, and the wrong one still prints "clean". The denylist was deleted rather than kept as a second opinion. The floor is now:

- **declared once** in `ruff.toml`'s `[per-file-target-version]` for `plugins/dev-team/**` — scoped deliberately, since repo-root `scripts/` and `tests/` run on a maintainer's interpreter and sit outside ADR 0014's scope;
- **enforced** by `chk_python_floor` in `scripts/ci-local.sh`, which resolves a real 3.8 and makes it byte-compile and import every shipped module via the new `scripts/import_probe_shipped.py`. It runs in the default gate, so the pre-push hook enforces it on every push, and it **fails rather than skips** when no 3.8 is reachable;
- **provisioned** by `scripts/dev-setup.sh` (via `uv`, since 3.8 is EOL), so that hard failure never strands a contributor;
- **pinned** by `tests/repo/test_python_floor.py`, which ties the declaration, the ADR, and the check together and asserts the check can never report "skipped".

Both new gates were made to fail on purpose before being trusted — the floor check twice, once with 3.9-only syntax planted in the shipped tree and once with the interpreter hidden from `PATH`.

**One follow-up needs `workflow` scope**, which this session's credentials lack (`git push` and the GitHub API both reject writes to `.github/workflows/`): add `chk_python_floor` to a job's `--only=` list in `plugin-tests.yml` so the floor is checked in CI as well as pre-push. One line.

## The rule behind all of it

Recorded in `CLAUDE.md` and in `agents/orchestrator.md`, where dispatch decisions are actually made: **never dispatch an agent or skill for work a deterministic tool can decide.** Tests, compilers, linters, parsers, validators and `git` answer mechanical questions; agents answer questions of judgement. A model aimed at a mechanical question returns a guess shaped like a result — silently wrong, confidently, and more expensive than the tool. Corollaries: verify a runtime property at runtime, and make every new gate fail once on purpose before trusting it.

## Test Plan

- [x] **Full `scripts/ci-local.sh` green — 5952 passed**, every check running (eslint included for the first time in this environment)
- [x] Shipped tree byte-compiles, imports, and passes all **2501 plugin tests on a real Python 3.8.20**
- [x] 183 new tests across the six slices, covering the issues' named acceptance criteria
- [x] 13 of those pin the five normalization collisions the security review found, verified against real `git diff` output
- [x] Both new gates demonstrated failing on purpose before being trusted
- [x] `python3 scripts/check_md_references.py` clean; knowledge index rebuilt

## Other deviations worth a reviewer's attention

- **#1627 whitespace allowlist.** The issue says "collapse changed-line pairs that differ only in whitespace". Implemented as written that is unsound for indentation-significant languages — dedenting a Python line moves it out of its block, and both sides strip identically. Collapsing is therefore gated on a brace-delimited extension allowlist, with unknown and extensionless files byte-exact, and quote-bearing lines always byte-exact. The security review confirmed this deviation was the correct call.
- **#1628 frontmatter vs. body.** The issue specifies `verify_model:`/`verify_effort:` *frontmatter*; implemented as **body** declarations, because #1333 reserves agent frontmatter for the official sub-agent contract and `agent-contract.json` is a dated upstream snapshot, not a schema this plugin may extend. Frontmatter would also make a typo invisible, since unknown keys are silently ignored.
- **#1629 changeset scoping.** All three new checks are changeset-scoped. `evals/README.md` forbids retrofitting the ~140 existing fixtures, and corpus-wide these emit ~154 findings. `--all` exists for deliberate triage.
- **The executable-claims convention validated itself while being written.** The first draft of the `evals/README.md` example asserted that `Intl.NumberFormat.prototype.format` returns `1` when read off the prototype. The probe disproved it in about a second: `format` is an accessor whose getter throws `TypeError` on property access, so it can never be extracted from the prototype at all. The committed example records the corrected probe, its output, and the Node version.

Closes #1624
Closes #1625
Closes #1626
Closes #1627
Closes #1628
Closes #1629
Part of #1623

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cm8KpYzgApt86ioqUePJv7)_


bdfinst:
## Adversarial security review — slice #1627 (required before merge)

Pass against the #1461 reviewers' checklist, attacking the claims listed in the PR body. Fixes and regression tests in b66c495.

**Verdict: the architectural claim holds; the implementation of it did not.**

The load-bearing argument is sound — the carry-forward exemption really is a property of content the hook recomputes at gate time from `git diff --cached`, never a claim written by the gated party, and the `>= 2` distinct-dispatch floor and the 1800s window really are untouched. I could not break that framing.

But "recomputed by the hook" only constrains anything if the recomputed function separates behaviorally different changesets. v1's normalization did not, in **five distinct ways**. One of them was a complete end-to-end reopening of #1461.

### Findings

| # | Severity | Finding |
|---|---|---|
| 1 | **Critical** | `sha256("")` was a universal carry-forward key |
| 2 | **High** | Parser confusion: a removed `-- ` line masqueraded as a file header |
| 3 | **High** | Changes with no hunk body were invisible to the digest |
| 4 | Medium | Canonical form was position-blind |
| 5 | Medium | A line moved between hunks read as formatting |
| 6 | Low | HTML-family files were treated as whitespace-insignificant |

---

**1 — `sha256("")` was a universal carry-forward key (critical).** `normalize_patch` returns `""` for a clean index as well as for any fully-cosmetic changeset, and `normalized_gate_hash` digested it anyway. Verified against real git: a clean index and a chmod-only stage both produced `e3b0c44298fc…`.

The full bypass chain needed no forgery at any step:

1. Two genuine review dispatches are recorded at any moment when nothing is staged — the ledger stamps `subject_hash_normalized = sha256("")` for both.
2. Stage a changeset that normalizes to empty (mode flip, rename, binary swap — findings 3–5 supply several).
3. The gate's raw lens rejects, the carry-forward lens recomputes `sha256("")`, matches, finds its two distinct dispatches, and passes with a `cosmetic-delta-carry-forward` event.

Two agents that reviewed nothing at all cleared the `>= 2` floor. `normalized_gate_hash` now returns `None` for an empty normalization, for exactly the reason it already did for a git failure: this lens means something only when two parties independently arrive at the same **non-trivial** value. Nothing real is lost — a wholly doc-classified changeset is already served by the doc-only exemption lens, which re-derives its own predicate.

**2 — Parser confusion via line prefixes (high).** A **removed** source line beginning `-- ` — a SQL, Lua, or Haskell comment — renders as `--- ...` in the diff. The parser dispatched on that prefix, read it as a file header, and set `in_hunk = False`, so every following line in that hunk vanished from the digest. Injected code included:

```sql
 CREATE TABLE t (id INT);
--- legacy note
+GRANT ALL PRIVILEGES ON *.* TO 'attacker'@'%';   ← invisible to the digest
 SELECT 1;
```

Added lines beginning `++ ` did the same through `+++ `. Hunk bodies are now consumed by the line counts in their own `@@` headers; a body that contradicts its header fails closed.

**3 — Changes with no hunk body were invisible (high).** `chmod +x`, a rename, a binary-blob swap, and an empty new file all produce a diff with no `+`/`-` lines at all, so staging any of them on top of already-corroborated content left the digest untouched. The in-code rationale — *"a mode change is real, but it is carried by the raw-hash lens, which is still evaluated first and still authoritative"* — does not hold: this lens runs **precisely when the raw lens has already rejected**, so it overrides that authority rather than deferring to it. Patch metadata is now recorded; binary files bind their `index` blob SHAs, the only content signal a textual diff exposes.

**4 — Position-blind canonical form (medium).** Flat per-file removed/added lists with hunk headers excluded gave one digest for the same line inserted at two different places (`+audit()` before vs. after a call). Each hunk's whole old side is now compared against its whole new side, **context lines included** — context is what makes an insertion point part of the subject. Note that hunk *line numbers* would not have fixed this: both variants carry an identical `@@` header.

**5 — Cross-hunk move read as formatting (medium).** Whole-file removed-equals-added dropped the file when a line was merely relocated between hunks — a `lock.acquire()` moved across a function. The collapse is now per hunk, never across hunks.

**6 — HTML-family whitespace (low).** `.html`/`.htm`/`.vue`/`.svelte` were in the whitespace-insignificant set, but `<pre>` and `<textarea>` preserve whitespace and the markup is not brace-delimited. Moved to the indent-significant set.

Also hardened: canonical fields are length-prefixed, so a source line containing a separator byte cannot forge a field boundary (git treats any NUL-free blob as text, so "no source line contains `\x1f`" was an assumption about attacker-supplied content).

### What held up under attack

- The `>= 2` distinct-dispatch floor and the 1800s recency window — genuinely untouched; re-verified that a single dispatch still blocks.
- Doc classification reusing the gate's own STRICT classifier: `agents/`, `skills/`, `.claude/`, `CLAUDE.md` can never be dropped as documentation.
- Quote-bearing lines compared byte-exactly — no string-literal edit rides the carry-forward.
- Indentation-significant extensions byte-exact, with unknown and extensionless files defaulting to byte-exact.
- Ledger events remain hook-computed and never self-asserted; events without `subject_hash_normalized` never match; a 1-line `.review-passed` can never carry forward.
- The verdict path fails closed on every error.
- The whitespace-allowlist deviation from the issue's literal text is correct as argued — implemented literally it would have let a Python dedent ride the carry-forward.

### Residual, recorded not fixed

The quote-character rule cannot see an **unquoted heredoc body** (Ruby `<<~SQL`, PHP `<<<TXT`, Perl `<<EOF`); reindenting a line inside one is a data change this module reads as formatting. It needs the same language awareness that defers comment stripping to v2, so it is documented in the module rather than papered over with a heuristic.

### Cost of the fixes

Findings 4 and 5 cost some invariance: a whitespace fix close enough to a reviewed change that git merges their hunks now shifts the digest, and the carry-forward is lost. That is the correct direction for this trade — a lost carry-forward costs one re-dispatch, a spurious one admits unreviewed code. The primary cases the slice exists for are unaffected, and are pinned against real git: reindenting the reviewed line itself, adding a markdown file alongside reviewed code, and rewriting doc content alongside reviewed code all still carry forward.

### Verification

All seven attacks blocked and all three invariance properties preserved against real `git diff` output, plus 13 new regression tests. Full local gate: **5939 passed, 7 skipped**. eslint could not run in this container (repo requires Node ≥ 24, container has v22.22.2, so `node_modules` never installed) — this branch changes no JavaScript, and CI covers it.

---

Issue #1638, opened 2026-07-31
fix(hooks): cosmetic-delta normalizer reads a reindented heredoc body as formatting


bdfinst:
Deferred from #1631 (slice #1627). Recorded in-module as `KNOWN RESIDUAL` in `plugins/dev-team/hooks/lib/review_gate_normalized_hash.py`.

## Context

#1627 added a normalization-invariant gate hash so review corroboration can carry forward across a re-stage that provably changed no behavior. What keeps *string data* out of the "cosmetic" bucket is a quote-character rule in `_canonical_line`: a changed line containing a quote character is compared byte-exactly, never whitespace-collapsed. That is what stops a whitespace-looking edit to a string literal from riding the carry-forward.

The adversarial security review of #1627 (five collisions found and fixed, one critical) confirmed the rule holds for quoted strings. It does not cover **unquoted heredoc bodies**.

## The gap

```ruby
sql = <<~SQL
  SELECT * FROM users WHERE role = admin
SQL
```

```php
$msg = <<<TXT
    Dear customer, your balance is overdue
TXT;
```

```perl
my $t = <<EOF;
  literal text that matters
EOF
```

None of those body lines contain a quote character, so `_canonical_line` treats them as ordinary code. Reindenting one is a **data change** this module reads as formatting — and if the file's extension is in the whitespace-insignificant (brace-delimited) allowlist, the change collapses and the carry-forward survives.

## Severity, honestly

Narrower than the indentation-significance hazard the allowlist already guards against, which moves code between blocks. The exposure needs all of: a heredoc in a whitespace-insignificant-extension file, an unquoted delimiter, a body whose leading whitespace is semantically load-bearing (SQL is mostly forgiving; templates and fixtures are not), and an attacker who controls the re-stage. It is a real hole, not a theoretical one, but it is not the `sha256("")` class of defect.

It was deliberately **not** papered over with a heuristic, because a regex that guesses at heredoc starts would produce a rule nobody can reason about, and false confidence here is worse than a documented gap.

## Fix

Needs the same **language awareness** that defers comment stripping to v2: knowing where a heredoc body begins and ends requires per-language lexing, not line-local inspection. Options, roughly in increasing cost:

[...]

PR #1677, opened 2026-08-01
fix(hooks): harden the review-gate multi-line-literal normalizer

### LINKED ISSUE #1660: fix(hooks): heredoc grammars for Lua long brackets, SQL dollar-quoting, XML CDATA, PHP inline HTML
## Context

Follow-up from #1638's review (found by arch-review, domain-review, security-review independently).

\`plugins/dev-team/hooks/lib/review_gate_normalized_hash.py\`'s \`_WHITESPACE_INSIGNIFICANT_EXTENSIONS\` opts \`.lua\`, \`.sql\`, \`.xml\`, and \`.php\` into whitespace collapsing, but none of them has a \`_HEREDOC_GRAMMARS\` entry for their own unquoted, whitespace-preserving multi-line literal forms:

- Lua long brackets: \`[[ ... ]]\` / \`[==[ ... ]==]\`
- PostgreSQL dollar-quoting: \`\$\$ ... \$\$\` / \`\$tag\$ ... \$tag\$\`
- XML element text / CDATA sections
- PHP inline HTML outside \`<?php ... ?>\`

Reindenting a line inside any of these is a real data change that currently normalizes away as "cosmetic" — the exact hazard #1638 closed for Ruby/PHP/Perl heredocs, still open for these forms.

## Suggested fix

Add grammar entries to \`_HEREDOC_GRAMMARS\` for Lua long brackets (level-counted open/close) and SQL dollar-quoting (tag-matched open/close). For XML/PHP-inline-HTML, either add a grammar or (fail-closed, cheaper) drop those extensions from \`_WHITESPACE_INSIGNIFICANT_EXTENSIONS\` — costs only an extra re-dispatch.

Whichever is chosen, restore an explicit "KNOWN RESIDUAL" note in the module docstring — #1638 replaced the original residual comment with the new grammar dict, so this class is no longer recorded as open anywhere in the file.

### LINKED ISSUE #1661: fix(hooks): heredoc-normalizer coverage gap for Go/JS/TS/C++ raw-string literals
## Context

Follow-up from #1638's review (found by correctness-review's final confirmation pass).

The same unquoted-multi-line-string-data hazard #1638 closed for Ruby/PHP/Perl heredocs also applies to other whitespace-insignificant languages with quote-free multi-line literal forms:

- Go backtick raw strings: `` `...` ``
- JS/TS template literals: `` `...` `` (only unsafe when interpolation contains no quote char on interior lines)
- C++ raw string literals: `R"(...)"`

\`.go\`, \`.js\`/\`.jsx\`/\`.ts\`/\`.tsx\`, and \`.cpp\`/\`.cc\`/\`.cxx\` are all in \`_WHITESPACE_INSIGNIFICANT_EXTENSIONS\` (\`plugins/dev-team/hooks/lib/review_gate_normalized_hash.py\`) with no \`_HEREDOC_GRAMMARS\` entry, so \`_heredoc_body_marks\` returns all-\`False\` and a reindent inside one of these literals normalizes away as cosmetic.

Note: #1638 replaced the module's original "KNOWN RESIDUAL (#1631 review)" comment (which used to record this whole class as open) with the new heredoc-grammar dict — so after #1638 merges, nothing in the file documents that this broader class is still partly open. At minimum, restore an explicit residual note; ideally add grammars for these forms too.

## Suggested fix

Add grammar entries (backtick-delimited for Go/JS/TS, \`R"delim(...)delim"\` for C++) to \`_HEREDOC_GRAMMARS\`, or document the residual explicitly in the module docstring if deferred.

### LINKED ISSUE #1662: fix(hooks): heredoc normalizer fails open beyond --unified=100000 for files over 100k lines
## Context

Follow-up from #1638's review (found by arch-review and security-review independently).

\`normalized_gate_hash()\` in \`plugins/dev-team/hooks/lib/review_gate_normalized_hash.py\` requests \`git diff --unified=100000\` so a heredoc opener anywhere earlier in a file is always part of the same hunk \`_heredoc_body_marks\` reasons over. This is a magic-number assumption ("any realistically-sized file"), not a verified precondition: a file with more than ~100,000 lines can still produce a hunk that does not start at line 1, making an opener above it invisible — the exact hazard the heredoc handling exists to close, reopened silently. Fails OPEN, not closed — the one exception to this module's stated fail-closed posture.

The data needed to detect this is already parsed and discarded: \`_parse_hunk_header\` captures \`old_start\`/\`new_start\` from the \`@@ -a,b +c,d @@\` header and throws them away (per a comment explaining why they were unused *before* #1638 — that reasoning is now stale for heredoc-grammar extensions).

## Suggested fix

Have \`_parse_hunk_header\` return the start lines. In \`normalize_patch\`, when a section's extension has a \`_HEREDOC_GRAMMARS\` entry and a hunk does not begin at old-line 1 / new-line 1, mark that hunk's lines byte-exact (or return \`None\`) rather than trusting \`--unified=100000\` blindly. This removes the dependency on the magic number and makes the invariant self-verifying for any caller, not just \`normalized_gate_hash\`.

### LINKED ISSUE #1663: fix(hooks): bound heredoc-normalizer — --unified=100000 has no size cap or subprocess timeout
## Context

Follow-up from #1638's review (found by performance-review, arch-review, and security-review independently).

\`normalized_gate_hash()\`'s \`--unified=100000\` change (\`plugins/dev-team/hooks/lib/review_gate_normalized_hash.py\`) turns a small diff into a whole-file diff for any changed file with a whitespace-insignificant extension. This runs synchronously in the \`pre_commit_review.py\` PreToolUse hook, with:

- No subprocess timeout on the underlying \`git diff\` call (via \`git_safe_diff.run_safe_git_diff\`).
- No cap on the resulting patch size before it's decoded and processed (\`normalize_patch\` holds several full in-memory copies: split lines, per-hunk old/new sides, canonicalized sides, joined strings, encoded records).

A one-line edit to a large tracked file (e.g. a big JSON lockfile or generated fixture — \`.json\` is in \`_WHITESPACE_INSIGNIFICANT_EXTENSIONS\`) now produces and processes that file's entire content twice, uncapped, on every commit.

## Suggested fix

- Add an explicit \`timeout=\` to the \`git diff\` subprocess call, treating \`TimeoutExpired\` as \`None\` (fail-closed, consistent with the module's posture).
- Cap the total patch size read by \`normalized_gate_hash\` (e.g. a few MiB), returning \`None\` above the cap rather than processing an unbounded payload.
- Set an explicit, conservative \`timeout\` on the \`pre_commit_review.py\` hook's own registration in \`plugins/dev-team/settings.json\` so its budget is stated rather than inherited from the harness default.

### LINKED ISSUE #1664: docs(hooks): telemetry-schema.md not updated for #1638's digest-algorithm change
## Context

Follow-up from #1638's review (found by arch-review).

\`normalized_gate_hash()\`'s digest algorithm changed in #1638 (heredoc-body marking + \`--unified=100000\` both change what bytes get canonicalized/hashed for affected files). \`subject_hash_normalized\` is written into \`.claude/metrics/boundary-events.jsonl\` by \`agent_dispatch_ledger.py\` at dispatch time and compared at commit time by \`pre_commit_review.py\`'s corroboration logic.

\`plugins/dev-team/knowledge/telemetry-schema.md\`'s \`subject_hash_normalized\` row already carries a compatibility note for the pre-#1627 case (when the field didn't exist at all) but was not updated to note that #1638 changed the digest's actual VALUE for any changeset touching a heredoc-grammar extension — events recorded by an earlier plugin version will never match after upgrade (fails closed — one lost carry-forward — not a correctness bug, but undocumented).

## Suggested fix

Add a line to \`telemetry-schema.md\`'s \`subject_hash_normalized\` row noting that #1638 changed the digest algorithm, mirroring the existing pre-#1627 note. Optionally: make the canonical form version-tagged (prefix the encoded record with an algorithm version) so future changes are self-describing rather than silently non-matching.

### LINKED ISSUE #1665: fix(hooks): heredoc-normalizer nits — mutable grammar dict, positional capture groups, naming
## Context

Follow-up from #1638's review. Three low-priority maintainability nits in \`plugins/dev-team/hooks/lib/review_gate_normalized_hash.py\`, none behavioral:

1. **\`_HEREDOC_GRAMMARS\` is a mutable module-level dict** (mutated after construction via \`_HEREDOC_GRAMMARS[".pm"] = ...\` / \`[".tf"] = [".hcl"] = ...\`), while its sibling language tables in the same module (\`_INDENT_SIGNIFICANT_EXTENSIONS\`, \`_WHITESPACE_INSIGNIFICANT_EXTENSIONS\`) are \`frozenset\`. Consider building the mapping in one literal and wrapping it in \`types.MappingProxyType\` (stdlib-only, 3.8-safe) so this safety-relevant table can't be mutated by an importer at runtime.

2. **Positional capture-group coupling.** \`_heredoc_body_marks\` unpacks every grammar's regex match via \`match.group(2)\` (identifier) / \`match.group(1)\` (modifier) — a convention documented only in a comment, not enforced by the type system. PHP's regex needs a phantom empty capture group \`()\` purely to keep this numbering aligned. Consider named groups (\`(?P<modifier>...)\`, \`(?P<ident>...)\`) so a future grammar author can't silently break the indexing, and PHP's regex could drop the empty-group workaround.

3. **Naming inconsistencies** (flagged by naming-review): the \`_extension()\` function's return value is stored as \`suffix\` everywhere it's used (\`_whitespace_collapsible\`, \`normalize_patch\`, \`_heredoc_body_marks\`'s parameter) rather than \`extension\`, inconsistent with the function's own name and the existing \`_..._EXTENSIONS\` constants' vocabulary. Also, \`old_heredoc\`/\`new_heredoc\` locals in \`normalize_patch\`'s hunk loop hold boolean MARK lists (from \`_heredoc_body_marks\`), not heredoc content — \`old_heredoc_marks\`/\`new_heredoc_marks\` would read more precisely.

All three are non-blocking style/robustness improvements, safe to batch into one PR.

### LINKED ISSUE #1666: fix(hooks): collapse near-duplicate per-language heredoc reindent tests into one parametrize
## Context

Follow-up from #1638's review (found by test-smell-review).

\`plugins/dev-team/tests/hooks/test_review_gate_normalized_hash.py\`'s \`TestHeredocBodies\` class has several near-identical per-language "reindent survives" tests (Ruby squiggly/dash/bare, PHP, Perl variants) sharing the same arrange-act-assert shape — build a before/after heredoc patch via \`_patch\`, assert \`normalize_patch(before) != normalize_patch(after)\` (plus a non-empty check on some but not all) — differing only in extension and opener/closer literal strings. The class already uses \`@pytest.mark.parametrize\` elsewhere in the same diff for an analogous case (the shovel-operator false-positive test), so the inconsistency is a test-organization gap rather than a missing tool.

Copies have already drifted within the same PR: some variants assert \`normalize_patch(after) not in ("", None)\` and others don't.

## Suggested fix

Collapse the near-duplicate reindent-survival tests into one \`@pytest.mark.parametrize("path,opener,closer,body_indent,body", [...])\`, keeping distinct tests only for cases with a genuinely different assertion shape (e.g. the \`.pm\`-shares-\`.pl\` identity test, the end-to-end git test). Low priority — maintainability only, no behavioral risk.

### LINKED ISSUE #1667: fix(hooks): heredoc-grammar misses — Ruby backtick, Perl empty-delimiter, quoted delimiters
## Context

Follow-up from #1638's review (found by security-review's and correctness-review's final confirmation passes). Three narrow, rare heredoc-opener forms are not matched by \`_HEREDOC_GRAMMARS\` in \`plugins/dev-team/hooks/lib/review_gate_normalized_hash.py\` — all in the unsafe (under-detection/missed-opener) direction, meaning a body reindent for these forms currently normalizes away as cosmetic:

1. **Ruby's backtick command heredoc**: `` result = <<`EOC` `` / `` <<~`EOC` `` (shell-command heredoc, documented in Ruby's Literals docs). The \`.rb\` opener's marker class \`['\"]?\` omits the backtick, so this form isn't recognized as an opener at all.
2. **Perl's empty-delimiter form**: \`print <<"";\` (blank-line-terminated heredoc). \`[A-Za-z_]\` can't match the empty string, so this opener is missed too.
3. **Quoted non-identifier heredoc delimiters**: Ruby/Perl technically permit an arbitrary quoted terminator (\`<<~'1SQL'\`, \`<<'9EOF'\`) that doesn't start with a letter/underscore — all three grammars require \`[A-Za-z_]\` as the first delimiter character, so a numeric-leading quoted delimiter is missed.

Per the module's own stated bias ("erring toward MORE lines counting as heredoc body, never fewer"), over-matching any of these is provably safe (marking only ever replaces a line with itself, byte-exact — never creates a digest collision), so widening is the correct direction whenever addressed.

## Suggested fix

- Add a backtick to the \`.rb\` opener's optional marker/trailer classes.
- Allow an empty identifier match for \`.pl\`'s blank-delimiter form (guard against ambiguity with a real word boundary).
- Broaden the quoted-delimiter branch of each grammar to accept a non-identifier-shaped body inside the quotes (e.g. \`[^'\"\s]+\`) when a quote/backslash marker is present.

Low priority — all three are rare in practice and were caught only by adversarial review, not real-world incidents.

### LINKED ISSUE #1676: fix(ci): chk_ruff and chk_eslint never run in CI, and ruff is unpinned
## Context

Found while working #1660–#1667. `scripts/ci-local.sh` declares `chk_ruff` and
`chk_eslint` in its **always-run** `CHECKS` array, so `bash scripts/ci-local.sh`
runs both locally. CI never does.

Every CI job delegates to `ci-local.sh` with an explicit `--only=` list, and no
list names either linter. Mechanically:

```
# union of every --only= in .github/workflows/, vs. the full CHECKS array
chk_eslint          <- never runs in CI
chk_eval_semver     <- never runs in CI
chk_harness_scope   <- never runs in CI
chk_nav_integrity   <- (link-check.yml is a documented CI superset)
chk_oe_staleness    <- advisory by declaration
chk_ruff            <- never runs in CI
chk_sa_mcp_tools    <- never runs in CI
```

So the only thing standing between a lint regression and `main` is that every
contributor has the pre-push hook installed **and** has the same linter version
as everyone else. Neither holds.

## The version skew is already real

`requirements-dev.txt` pins `ruff>=0.3.1` — no upper bound. On this container
both install paths are present and disagree:

```
$ ruff --version              # dev-setup.sh / uv-installed binary on PATH
ruff 0.15.8
$ python3 -m ruff --version   # pip install -r requirements-dev.txt
ruff 0.16.1

$ ruff check .                # what chk_ruff actually invokes
All checks passed!
$ python3 -m ruff check .
Found 21 errors.
```

Identical tree, two verdicts, decided by which ruff resolves first. `mypy>=1.11`
carries the same unbounded shape.

## It has already let a regression land

20 of those 21 findings are in `scripts/verify_toolchain.py` — added by
**#1669**, merged. One more is in `.claude/format_python.py`:

| Rule | Count |
| --- | --- |
| `UP032` use f-string instead of `.format()` | 8 |
| `UP006` use `list`/`tuple` over `List`/`Tuple` | 7 |
| `EXE001` shebang present but file not executable | 2 |
| `UP035` deprecated `typing` import | 2 |
| `UP045` use `X \| None` | 1 |

None are behavioral, which is exactly why nothing caught them. The point is not
the 21 findings — it is that a shipped file went in unlinted and the gate
reported success.

This is the repo's own `CLAUDE.md` rule, third occurrence: *a gate that cannot
fail is worse than no gate — it reads as a guarantee and delivers none.* Prior
occurrences: `engines.node` at `>=24` while containers ran Node 22 (husky inert,
eslint skipped, "All local CI checks passed"), and the Python 3.8 floor denylist
(#1650).

## Ask

1. Add `chk_ruff` and `chk_eslint` to a CI job's `--only=` list so the linters
   run on every PR, not only on machines with a working pre-push hook.
2. Pin `ruff` and `mypy` to exact versions in `requirements-dev.txt` so the gate
   is deterministic, and make `dev-setup.sh` install that same pinned version
   rather than whatever `uv`/`pip` resolves independently.
3. Fix the 21 outstanding findings so the pin can be raised cleanly.
4. Triage the remaining four never-in-CI checks (`chk_eval_semver`,
   `chk_harness_scope`, `chk_sa_mcp_tools`, and confirm `chk_nav_integrity`'s
   `link-check.yml` superset claim) — either wire them in or move them to
   `OPTIONAL_CHECKS` so the always-run array stops overstating its own coverage.
5. Per this repo's own rule, make the new gate fail on purpose once before
   trusting it.

bdfinst:
## Summary

- **Eight #1638 follow-ups, one PR** (#1660–#1667). All land in
  `review_gate_normalized_hash.py` and its suite, and they are genuinely
  entangled: #1665's named-capture-group refactor rewrites the very regexes
  #1660/#1661/#1667 extend, so splitting them would mean authoring the new
  grammars twice.
- **Two real bugs, not just coverage.** `--unified=100000` was a magic number
  standing in for "every hunk spans the whole file" — a file longer than the
  window yields a hunk starting partway down with its heredoc opener
  invisible, failing **OPEN**, the one exception to this module's fail-closed
  posture (#1662). And the same call had no subprocess timeout and no payload
  cap, so a one-line edit to a large `.json` file canonicalized the whole file
  several in-memory copies deep, synchronously inside a PreToolUse hook, on
  every commit (#1663).
- **A separate CI defect found while working, filed as #1676 and fixed here.**
  `chk_ruff` and `chk_eslint` sat in `ci-local.sh`'s always-run array but were
  named by no CI job's `--only=` list, so both linters only ever ran on a
  contributor machine with a working pre-push hook. `scripts/verify_toolchain.py`
  reached `main` via #1669 carrying 20 ruff findings that way, and this
  container had two ruffs installed at once (0.15.8 on PATH, 0.16.1 from an
  unbounded `ruff>=0.3.1`) disagreeing by 21 findings on an unmodified tree.

## Test Plan

- [x] `python3 -m pytest plugins/dev-team/tests tests/repo tests/hooks -q` — 3337 passed
- [x] Full `bash scripts/ci-local.sh` — 6153 passed, 43 skipped, all checks green
- [x] New grammars verified against 16 hand-built cases (Lua levels, SQL tags,
      Go/JS/C++ raw strings, PHP inline HTML, Ruby backtick, Perl empty
      delimiter) plus the shovel-operator negatives (`arr << item`, `x << 2`)
- [x] `_HEREDOC_GRAMMARS` identity aliases (`.pm` is `.pl`, `.tf`/`.hcl` is
      `.rb`) preserved under `MappingProxyType`
- [x] Exercised on a real Python 3.8 interpreter — module imports and both new
      inline grammars return correct marks (not just byte-compiled)
- [x] **Gates made to fail on purpose before being trusted**, per this repo's own
      rule: an unused import drives `chk_ruff` to exit 1 (and exit 0 clean);
      loosening the pin fails `test_ruff_is_pinned_exactly`. The #1662 guard is
      self-demonstrating — `TestHunkStartGuard`'s first two tests differ *only*
      in hunk start line and assert opposite outcomes.
- [x] New `Lint (ruff + eslint)` CI job confirmed **green on this PR** — the
      first CI job to run `npm ci`, so eslint has somewhere to live at all.

## Notes for review

- `.xml` is **dropped** from `_WHITESPACE_INSIGNIFICANT_EXTENSIONS` rather than
  given a grammar. Element text and CDATA preserve whitespace across most of a
  document, so a grammar would mark nearly every line anyway — #1660 offers this
  as the cheaper fail-closed option, and it costs one re-dispatch, never gate
  strength.
- PHP's *leading* inline-HTML region (before a file's first `<?php`) is a
  recorded KNOWN RESIDUAL rather than a fix; it needs a `?>` to open. #1638 had
  displaced the module's original residual note, so this class was documented
  nowhere — it is restored.
- Every over-match added here is safe by construction: marking replaces a line
  with itself, byte-exact, so it can only cost a carry-forward, never create a
  collision. Under-matching is the unsafe direction, and that is the one these
  grammars close.
- The `.github/workflows/` half of #1676 was pushed by @bdfinst directly — this
  session's credentials are refused for workflow files without `workflow` scope.

Closes #1660
Closes #1661
Closes #1662
Closes #1663
Closes #1664
Closes #1665
Closes #1666
Closes #1667
Closes #1676


bdfinst:
CI red on **Plugin content & hooks** — my error, root-caused and fixed, but the fix touches `.github/workflows/` so I can't push it (this session's credentials lack `workflow` scope). Patch handed to @bdfinst to apply.

**What broke.** I put `chk_harness_scope` in the content-guard job. Its checks 5–6 shell out to the red-team `orchestrator.py`, which imports `httpx` via `lib/http_client.py` and `lib/scoring.py`. That job installs only `jq` + `pytest` + `pytest-asyncio`, so both end-to-end checks die on import.

I'd verified the script had no *direct* third-party imports and stopped there — the dependency is one subprocess hop away, which a static import check doesn't see. Same shape as the thing #1676 is about, which is a little pointed.

**Reproduced locally** by shadowing `httpx` on `PYTHONPATH`:

```
ImportError: simulated missing httpx
=== FAILED: 2 check(s) ===
```

Two failing checks, matching CI exactly.

**Fix.** Move it to the `harness-smoke` job, which already installs `httpx` and already runs the sibling `chk_harness_smoke`:

```diff
-          --only=...,chk_hook_units,chk_sa_mcp_tools,chk_harness_scope
+          --only=...,chk_hook_units,chk_sa_mcp_tools
...
-        run: bash scripts/ci-local.sh --only=chk_harness_smoke
+        run: bash scripts/ci-local.sh --only=chk_harness_smoke,chk_harness_scope
```

Both placements verified locally: `--only=chk_harness_smoke,chk_harness_scope` passes, and the reduced content-job set passes. `chk_sa_mcp_tools` stays in the content job — it really is stdlib-only.

Nothing else on the PR is affected; every other job is green, including the new `Lint (ruff + eslint)`.


---

Issue #1735, opened 2026-08-03
feat(repo-review): add a repo-wide review skill for drift-style agents removed from per-diff code-review


bdfinst:
## Problem

#1733 proposes removing `token-efficiency-review` and `claude-setup-review` from `/code-review`'s per-diff roster (both already grouped in `EXEMPT_AGENTS` as pure config/frontmatter validators of repo-wide state, not diff semantics). Per further discussion, `ai-provenance-review` is also confirmed for removal — it flags accumulating "verification debt," a trend metric, not a per-diff correctness gate.

Removing them from per-diff dispatch only makes sense if something else actually runs them — otherwise these checks just stop happening. This issue is that something else: a `/repo-review` skill, structurally similar to `/code-review`, that runs a whole-repository pass instead of a diff-scoped one.

## Proposed shape

- **Mirrors `/code-review`'s dispatch/aggregation machinery** (agent panel, `Context needs` contract, health scoring, report output) but scoped to the whole tree rather than a changeset — no `git diff`/staging/gate-hash machinery, since there's no `.review-passed` gate to satisfy; this is a report, not a commit gate.
- **Roster**: the three agents removed from `/code-review`'s per-diff panel —
  - `token-efficiency-review` (file length, CLAUDE.md size, LLM anti-patterns — needs whole-tree visibility to see cross-PR creep at all)
  - `claude-setup-review` (CLAUDE.md completeness, agent frontmatter schema compliance)
  - `ai-provenance-review` (verification-debt sweep across the whole codebase, not just recently-touched files)
  - plus `component-architecture-review` **unconditionally** here (cross-file duplication/reusable-component extraction is best judged against the *entire* component inventory — see #1733's discussion). This one is NOT removed from `/code-review` — see the per-diff rule below.
- **Cadence**: triggered intermittently, not on every commit — volume-based (e.g. every N merged PRs) is the better fit than calendar time for these agents specifically, per #1733's reasoning. Exact mechanism (state file tracking last-audit point? scheduled GitHub Action? manual `/repo-review` invocation only?) is a design decision for the spec, not fixed here.
- **Output**: a report artifact (e.g. `.dev-team-reports/repo-review.md`) or a persistent tracking issue that gets updated each run, showing drift trend over time — not a pass/fail gate blocking anything.

## `component-architecture-review`'s dual placement

Per #1733 discussion, this agent's rule is more nuanced than a straight removal:

[...]

PR #1744, opened 2026-08-03
fix(hooks): repo_review_nudge should trigger on % of codebase changed

### LINKED ISSUE #1743: fix(hooks): repo_review_nudge should trigger on % of codebase changed, not an absolute line count
`repo_review_nudge.py` (#1739/#1740) currently nudges once added lines since the last `/repo-review` run cross a fixed absolute threshold (`DEV_TEAM_REPO_REVIEW_LINE_THRESHOLD`, default 2000). An absolute line count doesn't scale with repo size — 2000 lines is noisy drift-signal for a 5k-line repo and meaningless for a 3M-line one. It should trigger on the **percentage of the codebase that has changed** since the last run instead.

## Fix

- New signal: `added_lines_since_baseline / total_current_loc * 100`.
  - `total_current_loc`: `git diff --numstat <empty-tree>..HEAD` summed (everything in HEAD is necessarily an "addition" relative to the empty tree, so this doubles as a cheap, accurate current-LOC count — no need to shell out to `wc -l` across every tracked file).
  - `added_lines_since_baseline`: same `git diff --numstat` call, baseline = `last_commit` if on record and reachable, else the empty tree again (so "never run" naturally resolves to exactly 100% — the whole current codebase is undreviewed — rather than a special-cased branch).
- Rename `DEV_TEAM_REPO_REVIEW_LINE_THRESHOLD` → `DEV_TEAM_REPO_REVIEW_PERCENT_THRESHOLD`, default `1` (percent) — a starting guess, not measured, same honesty framing as the value it replaces.
- Nudge message reports the percentage (and the raw added/total counts for context).

## Also fold in: validate `last_commit` before using it in a git command

While updating this, `skills/repo-review/SKILL.md` (merged in #1742) added exactly the hardening this hook is missing: `.claude/memory/repo-review-state.json` is written by `/repo-review` but is untrusted input — nothing stops a hostile PR from committing its own copy with an arbitrary `last_commit` string (the file is meant to be gitignored, but a hostile actor can still force-add it). The skill now validates `last_commit` matches `^[0-9a-fA-F]{7,40}$` before it touches any git command; this hook reads the same file and passes `last_commit` straight into a `git diff` subprocess arg without that check. Not shell-interpolated (so classic shell injection doesn't apply), but an unvalidated value starting with `-` could still be misparsed as a git flag rather than a revspec (argument injection). Add the same regex validation here, treating a non-matching value exactly like "no prior state."

Part of #1735

bdfinst:
## Summary
- Replaces the absolute added-lines threshold in `repo_review_nudge.py` with a size-independent **percentage of the codebase changed**: `added_lines_since_baseline / total_current_loc * 100`.
- `total_current_loc` reuses the same added-lines counter, diffed against git's empty-tree object — no extra `wc -l` pass needed.
- "Never run" now resolves to exactly 100% (added == total) instead of a special-cased branch.
- Renames `DEV_TEAM_REPO_REVIEW_LINE_THRESHOLD` → `DEV_TEAM_REPO_REVIEW_PERCENT_THRESHOLD` (default `1`, meaning 1%).
- Also fixes a gap noticed while updating this: `last_commit` from `.claude/memory/repo-review-state.json` is untrusted input (the file is meant to be gitignored, but a hostile PR could still force-add one), and `skills/repo-review/SKILL.md` (merged in #1742) already validates it against `^[0-9a-fA-F]{7,40}$` before touching any git command. This hook read the same file without that check — added the identical validation.

## Test plan
- [x] `plugins/dev-team/tests/hooks/test_repo_review_nudge.py` — 10 tests, rewritten for percentage semantics (never-run == 100%, below/above threshold since last run, unreachable last_commit, invalid last_commit shape treated as never-run, malformed state file, invalid env var falls back to the 1% default)
- [x] `python3 -m ruff check` — clean
- [x] Full `plugins/dev-team/tests/hooks tests/hooks tests/repo` — 2945 passed, 46 skipped

Closes #1743
Part of #1735

### DIFF
--- a/plugins/dev-team/hooks/repo_review_nudge.py
+++ b/plugins/dev-team/hooks/repo_review_nudge.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""repo_review_nudge — Claude Code SessionStart hook (#1739/#1743).
+"""repo_review_nudge — Claude Code SessionStart hook (#1739/#1743/#1746).
 
 `/repo-review` (#1733/#1735) tracks drift in
 `.claude/memory/repo-review-state.json` (`{"last_commit", "last_run_at"}`)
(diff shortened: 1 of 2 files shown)


bdfinst:
Bumped the default threshold from 1% to 10% per feedback — still a starting guess, still overridable via `DEV_TEAM_REPO_REVIEW_PERCENT_THRESHOLD`.

bdfinst:
Added an absolute floor (default 200 added lines) and ceiling (default 5000) around the percent trigger — pure percentage was too naggy on small repos and could go silent on large ones. See #1746.

Issue #2031, opened 2026-08-26
perf(mutation-kill): confirm convergence with a survivor-scoped Stryker re-run


bdfinst:
## Problem

Every file that converges pays one **full** scoped mutation run to learn it has converged, and that run generates nothing.

`_score_round` (`plugins/dev-team/skills/mutation-testing/scripts/mutation_kill_loop.py`) opens each round with `run_scoped_stryker` over the file's entire mutant set, extracts survivors, then calls `stop_reason`. When the answer is "no improvement", the round ends there — the whole mutant set was re-mutated and re-tested to answer a question about the handful of mutants that survived the previous round.

The information is genuinely necessary — it is what proves convergence and produces the file's final honest score for the report. It is the *price* that is wrong, not the measurement.

## Why this is tractable here

Both halves already ship:

- **The spans are already computed.** `mutation_report.survivors_by_line()` (`mutation_report.py:408`) returns the previous round's Survived mutants clustered by source line. That is exactly the set the confirm-run needs to re-test.
- **The narrowed-config pattern is already established.** `stryker_timeout_retry.py:82` emits a config scoped to a subset for the timeout-retry path, and round-1 baseline reuse (#1544, #1545, extended to `--concurrency > 1` in #1920) is the same "don't re-run Stryker when the answer is already obtainable" idea at the other end of the loop.
- **The splice pattern is already established.** Phase 8 (`/quality-targets-converge`, #1208) reports a whole-repo score by splicing freshly-measured changed files over the persisted Phase-2 baseline. A survivor-scoped confirm result splices over the prior round's report the same way.

Today `mutate` is only ever set at **file-glob** granularity — `mutation_kill_loop.py:215` writes `"mutate": ["**/{source_file}"]`, and every reference example in `csharp-stryker-net.md` is a glob. Narrowing below that is the new part.

## Proposed change

1. On the terminal round, build a confirm-scoped config from the previous round's `survivors_by_line()` output rather than the file glob, and run that.
2. Splice the confirm result over the prior round's report to produce the file's final honest score, so the reported number keeps the same shape and denominator it has today.
[...]

Issue #1999, opened 2026-08-25
feat(harness): epic — dev-loop and test economics


bdfinst:
## Context

Epic #1973 optimized the `/code-review` panel — lens tiering plus deterministic offload, measured at a 12.1% reduction in lens dispatches. Session-report analysis since then shows that panel dispatch **count** was the smaller lever. Two independent extraction passes over the 26 Jul – 25 Aug 2026 corpus agree on the shape:

- **~90% of token spend is context**, not generation (cache read 57.1% + cache write 32.6% vs 10.3% generation).
- **Session length dominates**: 35 sessions out of 6,748 (0.5%) account for 50% of main-thread spend. The same turn costs ~3× at 900k context as at 100k.
- **Exploration call *count*, not payload, fills context**: ~20k grep/cat/find calls returning only 53 MB total. Each call re-sends the whole prefix.

This epic covers the cost of **doing development and improving tests** — the inner loop — as distinct from #1973's review-panel scope.

## Sub-issues

Filed as sub-issues of this epic. Ordering note: none of these depend on the review-contract fix, which is tracked separately as a correctness defect.

## Non-goals

- Review-panel composition and routing — that stays in #1973.
- The session-report extractor's own blind spots — #1990 / #1994.

## Standing constraint

Per `CLAUDE.md`: *deterministic tools over inference*. Every slice here should end in a program that computes an answer, not an agent that estimates one. And per #1515's finding, recorded in `skills/build/SKILL.md`: **fan-out never saves tokens, it trades them for wall-clock** — sequential is the correct default everywhere token cost is the objective.

bdfinst:
## Handoff — remaining priority clusters (4–7)

Posted on this epic because it is the durable home for the work; `.claude/memory/` is gitignored and the authoring session's container is ephemeral. Companion to the [session handoff](https://github.com/bdfinst/agentic-dev-team/pull/2053#issuecomment-5427128074) and its [correction](https://github.com/bdfinst/agentic-dev-team/pull/2053#issuecomment-5427188946) on #2053.

### Counts, corrected

An earlier note said "12 of 30". Both numbers were wrong. Verified against the `priority` label:

| | |
|---|---|
| Tagged `priority` | **34** |
| Closed | **2** (#2001, #2030 — merged in #2052) |
| Addressed, awaiting #2053 merge | **9** (#1953, #2027, #1950, #1952, #2033, #2003, #2002, #2005, and #2031 *partial*) |
[...]

Issue #1938, opened 2026-08-06
fix(mutation-kill): harden retry/downgrade mechanism


bdfinst:
## Consolidation

Groups three open hardening findings against the retry-then-downgrade mechanism (`mutation_kill_retry.py` / `stryker_shard_pipeline.py`, introduced by #1908) into one batch, per the issue-review at #1759.

- #1917 — gateway-class error classification matches unanchored substrings in truncated stderr
- #1918 — retry/downgrade state and policy are anemic and welded to the subprocess transport
- #1919 — generation-exhausted outcome is invisible past a log line in --all shard runs

All three surfaced from closing-pass review of Slice 3 (#1908) and touch the same module family, so they're best landed as one PR to avoid repeated review churn on the same file.

Closes #1917
Closes #1918
Closes #1919

Issue #1790, opened 2026-08-04
fix(test-improve): wire per-Story coverage deltas into Phase 5 as a live steering signal


bdfinst:
**Problem:** Per-Story coverage deltas (produced by the coverage-delta skill/worker) were only reviewed at the end of the nextgen-core Pass 1 run, not used as a live signal during Phase 5 to catch that coverage had gone flat while mutation-kill work proceeded.

**Failure scenario:** A multi-Story Phase 5 run keeps completing Stories and reporting mutation-kill progress while line/branch coverage stays flat for several Stories in a row; nothing surfaces this until the final report, by which point the phase's budget is spent on the wrong layer.

**Fix:** After each Story's coverage-delta measurement in Phase 5, compare against a minimum expected per-Story delta (or a running average) and flag/warn when several consecutive Stories show near-zero coverage movement, prompting a re-check of Phase 1 targeting (see #1785) before continuing.

Part of #1785

bdfinst:
**Proposed batch grouping (autoship)**

An automated grouping pass proposes shipping this issue together with #1786, #1787, #1790 as one batch.

Rationale: #1786, #1787, and #1790 all propose edits to the same file, plugins/dev-team/skills/test-improve/SKILL.md, at different phases (Phase 1 targeting, Phase 0 goal-conflict detection, Phase 5 live steering) of the same nextgen-core retro follow-up (#1785). #1787 explicitly references the Phase 1 fix from #1786 as a precondition. Shipping these as separate PRs would very likely produce merge conflicts on the same skill file and phase-ordering inconsistencies.

Member issues: #1786, #1787, #1790

This issue has been labeled `autoship:blocked` pending human confirmation. To confirm this batch and let it ship together on the next autoship round, run:

```
gh issue edit 1786 1787 1790 --add-label autoship:batch-confirmed --remove-label autoship:blocked --add-label autoship:ready
```

To reject: relabel individual issue(s) back to `autoship:ready` without adding `autoship:batch-confirmed` — they'll be re-evaluated solo on the next round.

PR #2053, opened 2026-08-26
fix(mutation-kill): four shard-pipeline and retry defects, plus yield steering

### LINKED ISSUE #1950: is_gateway_class_error misses common real-world 5xx shapes: underscore-joined status keys, status 529, stdout-only errors
## Context

During the slice-review checkpoint for #1938 Slice 1 (#1917 — typed `HeadlessCallFailed` + anchored gateway-class regex, `plugins/dev-team/skills/mutation-testing/scripts/mutation_kill_retry.py`), the `domain-review` and `correctness-review` agents independently identified real gaps in `is_gateway_class_error`'s classification of "gateway-class" (upstream 502/503/504-style transient) failures. These are genuine, verified gaps — not implementation bugs against #1917/#1938's own acceptance criteria, which were met exactly as specified and reviewed across a full 5-critic plan-review cycle plus a build-time acceptance-criteria gate. The exact regex (`r"\b(?:status|http|code)[^0-9]{0,12}50[234]\b"`) was a deliberate "decided now, not deferred" choice in the plan, not an oversight — but three real-world failure shapes still evade it:

1. **`\b` word-boundary rejects underscore-joined status keys.** `_` is a `\w` character, so `error_code: 503` or `errorCode=504` never match — `\b` requires a non-word character immediately before `status`/`http`/`code`, and `_c`/`rC` transitions are both word-to-word. Confirmed by direct regex reasoning (domain-review) and independently corroborated by correctness-review's S2 note about `HTTP/1.1 502` failing for the same digit-in-the-gap reason.
2. **Status 529 (Anthropic's own overload code) is not in the numeral alternation** (`50[234]`), and the downgrade ladder this policy drives is Anthropic-specific — so 529 is plausibly the single most common real "retry me" signal this classifier will ever see. It's covered today only via the `overloaded_error` marker string, which a rendered `API Error: 529 Overloaded` message won't contain verbatim.
3. **The classifier is structurally blind to stdout.** `run_claude_headless` raises `HeadlessCallFailed(result.returncode, result.stderr)` and discards `result.stdout` entirely (`mutation_kill_shared.py`). If `claude --print` ever renders its error payload on stdout with a non-zero exit, no amount of stderr-side regex hardening helps.

## Why deferred rather than fixed inline

Each of these requires an actual domain-modeling decision, not a mechanical fix:
- Widening the regex changes what "gateway-class" means and needs its own boundary tests (what should NOT match?).
- Adding 529 needs an explicit decision on whether the Cloudflare 52x family and bare 500 are in scope too.
- Scanning stdout needs a decision on whether `HeadlessCallFailed` should carry both streams (`stdout=''` field) and whether `__str__`'s message-format contract changes.

This mirrors the precedent set by #1948 (a similar deliberately-deferred gap found during #1937's review): confirmed real, confirmed safe to defer (a missed gateway-class classification means a file's retry-then-downgrade doesn't trigger and it just fails without retry — no downgrade-ladder corruption, no false-positive commit-trailer pollution), and filed as a follow-up rather than expanding an already-reviewed slice's scope mid-flight.

## Suggested fix shape (not binding)

- Anchor from the digit side instead of the word side, e.g. `(?<[\w.])5(?:0[234]|29)(?[\w.])` plus a proximity check for a status-context word within ~24 chars on either side — admits `error_code: 503`, `HTTP/1.1 502`, and `upstream returned 503 status` while still rejecting `request id 502391`.
- Add `"overloaded"` alongside `"overloaded_error"` in `_GATEWAY_CLASS_MARKERS`, and explicitly decide (and document) whether 500/52x are in scope.
- Carry `stdout` on `HeadlessCallFailed` (`HeadlessCallFailed(returncode, stderr, stdout='')`) and have `is_gateway_class_error` scan the concatenation, keeping `__str__` stderr-only so the existing message contract is unchanged.

## Acceptance criteria (draft — refine at planning time)

- [ ] `error_code: 503` (or another underscore/camelCase-joined status key) classifies as gateway-class
- [ ] A rendered `529`/`Overloaded` message classifies as gateway-class, with an explicit, documented decision on whether bare `500` and Cloudflare 52x codes are in scope
- [ ] A non-zero exit whose error payload appears only on stdout (not stderr) still classifies correctly when it should
- [ ] Existing negative cases (`request id 502391`, a plain non-gateway `RuntimeError`) still correctly classify as non-gateway-class
- [ ] `HeadlessCallFailed`'s module/class docstrings are updated to describe the (now more permissive) classification precisely — avoid re-introducing an "EXACT" over-claim

### LINKED ISSUE #1952: _format_downgrade_message misreports exhaustion reason for a "fable" fallback-model override
## Context

During the slice-review checkpoint for #1938 Slice 2 (#1918 — non-anemic `_RetryState`, injectable transport, derived `DowngradeEvent.exhausted`), `domain-review` independently identified a real, verified bug in `_format_downgrade_message` (`plugins/dev-team/skills/mutation-testing/scripts/mutation_kill_retry.py:160-180`) — a function Slice 2 did **not** touch (confirmed: no step in #1938's plan modifies it, and the `.exhausted` property's derivation is byte-identical in value to the old stored field it replaced, verified independently by `correctness-review` during the same checkpoint). This is pre-existing behavior from the original #1908 implementation, not introduced or worsened by #1938.

## The bug

`_format_downgrade_message` infers *why* generation exhausted from a proxy — `from_model in _KNOWN_LADDER_MODELS` (`_KNOWN_LADDER_MODELS = frozenset({"opus", "sonnet", "haiku"})`, line 159) — instead of using the actual reason the caller already knew and discarded:

```python
if event.exhausted:
    normalized_from = event.from_model.strip().lower() if event.from_model else None
    ...
    if normalized_from in _KNOWN_LADDER_MODELS:
        return (... "no further downgrade will be attempted")
    return (... "no downgrade ladder position available for this model, surfacing to operator")
```

`mutation_kill_shared._VALID_FALLBACK_MODELS = frozenset({"opus", "sonnet", "haiku", "fable"})` — `fable` is explicitly documented (line 321-323) as "a valid `--model`/`DEV_TEAM_MUTATION_FALLBACK_MODEL` override value but deliberately has no entry in `_MODEL_DOWNGRADE_LADDER`."

**Failure scenario:** a file starts on `opus`, downgrades once to `fable` (via `DEV_TEAM_MUTATION_FALLBACK_MODEL=fable`), then exhausts on a later gateway-class failure. At that point `event.from_model == "fable"`, which is **not** in `_KNOWN_LADDER_MODELS`. The function takes the "no downgrade ladder position available for this model" branch — but a downgrade demonstrably *did* happen (opus→fable). The operator-facing message claims no downgrade occurred when one did.

The root cause: `_retry_once_then_maybe_downgrade` (line ~281) already knows the real reason via its `already_downgraded` local (was this call's exhaustion because the ladder ran out, or because the one-time downgrade budget was already spent?) but discards it before constructing `DowngradeEvent`. `_format_downgrade_message` then has to re-guess the reason from an unrelated fact (ladder membership of the *current* model), which happens to line up for `opus`/`sonnet`/`haiku` but not for the `fable` override.

## Why deferred rather than fixed inline

`_format_downgrade_message` is entirely outside #1938 Slice 2's declared file scope (`mutation_kill_retry.py`/`test_mutation_kill_retry.py` were touched, but not this specific function's logic — only `DowngradeEvent`'s field-to-property conversion, which left this function's behavior unchanged). Fixing it properly needs a real modeling decision, not a one-line patch:

## Suggested fix shape (not binding)

Add an explicit exhaustion reason to `DowngradeEvent` — e.g. `reason: Literal["downgraded", "budget-spent", "no-ladder-position"]` — set from the branch that actually knows (`already_downgraded` vs. `resolve_fallback_model(...) is None`), keep `exhausted` as `reason != "downgraded"`, and switch `_format_downgrade_message` on `reason` instead of `_KNOWN_LADDER_MODELS` membership.

## Acceptance criteria (draft — refine at planning time)

- [ ] A file that downgrades to `fable` and later exhausts prints a message correctly attributing the exhaustion to the spent downgrade budget, not "no downgrade ladder position available"
- [ ] Existing messages for the `opus`/`sonnet`/`haiku` ladder-floor case and the never-downgraded no-ladder-position case are unchanged
- [ ] A regression test pins the `fable`-override exhaustion message specifically (none currently exists — confirmed via the #1938 Slice 2 test-review pass, which enumerated all 35 tests in `test_mutation_kill_retry.py` and found none exercising this path)

### LINKED ISSUE #1953: stryker_shard_pipeline.py: default git_run path crashes with TypeError on every real invocation
## Context

Found by `correctness-review` during the final `/pr`-gate review of #1938 (a hardening PR for an unrelated part of the mutation-kill pipeline — this bug predates and is untouched by that PR, confirmed via `git diff` showing zero changes to this region since before #1938's branch was cut).

## The bug

`plugins/dev-team/skills/mutation-testing/scripts/stryker_shard_pipeline.py`:

```python
def _git(cmd: Sequence[str], repo_root: Path, *, check: bool) -> subprocess.CompletedProcess:
    return subprocess.run(list(cmd), cwd=str(repo_root), capture_output=True, check=check)


GitRunner = Callable[[Sequence[str], Path, bool], object]


def worktree_add(repo_root: Path, path: Path, *, ref: str = "HEAD", git_run: GitRunner | None = None) -> None:
    git_run = git_run or _git
    if Path(path).exists():
        git_run(["git", "worktree", "remove", "--force", str(path)], repo_root, False)
    git_run(["git", "worktree", "add", "--quiet", str(path), ref], repo_root, True)


def worktree_remove(repo_root: Path, path: Path, *, git_run: GitRunner | None = None) -> None:
    git_run = git_run or _git
    git_run(["git", "worktree", "remove", "--force", str(path)], repo_root, False)
```

`_git` declares `check` as **keyword-only** (`*, check: bool`), but every call site passes it **positionally** — `worktree_add` (2 sites) and `worktree_remove` (1 site) all call `git_run([...], repo_root, False)` / `git_run([...], repo_root, True)`. The `GitRunner` type alias itself documents the intended (positional) signature: `Callable[[Sequence[str], Path, bool], object]`.

Since `git_run = git_run or _git`, **any real invocation that doesn't inject a fake `git_run`** — which is every production path, since `main()` calls `run_all(...)` with no `git_run` argument at all — falls through to `_git` and crashes:

```pycon
>>> def _git(cmd, repo_root, *, check): ...
>>> _git(["git", "worktree", "add"], "root", False)
TypeError: _git() takes 2 positional arguments but 3 were given
```

This means `worktree_add`, called from `process_shard` on the very first shard of any real run, crashes the entire pipeline before Stryker ever executes a single mutant.

## Why the test suite never caught it

Every test in `tests/scripts/test_stryker_shard_pipeline.py` injects its own 3-positional fake (`_Recorder.git_run(self, cmd, repo_root, check)`, line ~90) — matching `GitRunner`'s documented signature, not `_git`'s actual one. `_git` itself is never exercised by any test, so the mismatch is invisible to CI despite the suite being green.

## Fix

Either drop the `*` from `_git`'s signature (`def _git(cmd, repo_root, check: bool)`), or pass `check=` as a keyword at all three call sites. The former is the minimal fix and keeps `_git`'s signature consistent with `GitRunner`'s documented shape.

## Suggested acceptance criteria

- [ ] `_git`'s signature accepts `check` positionally, matching `GitRunner`'s type alias and all three existing call sites
- [ ] A test exercises `_git` itself (not just the `_Recorder` fake) so this class of signature drift is caught in future
- [ ] `worktree_add`/`worktree_remove` succeed end-to-end with no injected `git_run` (i.e. the real default path), verified against a real git repo in a test or manual run

### LINKED ISSUE #2027: fix(ci): lint-staged invokes the bare ruff binary, bypassing #1676's pinned-module rule
## Summary

`package.json`'s `lint-staged` config runs the **bare `ruff` binary from PATH**:

```json
"*.py": "ruff check --fix"
```

Every other ruff invocation in this repo deliberately does not. #1676 established the rule and `scripts/ci-local.sh` states it at the call site:

> Invoked as `python3 -m ruff`, never the bare `ruff` binary (#1676). Both can be present at once and disagree — this repo's own container carried a uv-installed 0.15.8 on PATH alongside requirements-dev.txt's 0.16.1, which reported 21 findings the binary called clean. The module form resolves to the version this repo **PINS**, so the gate means the same thing here, in CI, and on every contributor machine.

CI follows the same rule — `plugin-tests.yml` installs `"$(grep -E '^ruff==' requirements-dev.txt)"` and the job comment says *"this job and a contributor's machine cannot disagree about what 'clean' means."*

The pre-commit hook is the one invocation site that never got the fix. So the protection holds for `ci-local.sh` and CI, and is absent from the path that runs on **every single commit**.

## It recurred, in the opposite direction

Observed 2026-08-25 in this repo's own container, which still carries both:

| invocation | version | verdict on `repo_invariants.py` |
|---|---|---|
| `python3 -m ruff` (pinned, what CI runs) | 0.16.1 | clean |
| `ruff` from PATH (`/root/.local/bin/ruff`, what lint-staged runs) | 0.15.8 | **E402, 1 error** |

A commit was rejected by the pre-commit hook on a line the author never touched, after `ruff check` had been run and passed. Note the direction is inverted from #1676's case: there the pinned version was *stricter*, here the PATH version is. That is the point — two versions disagree per-rule, so neither "newer is stricter" nor "older is stricter" is a safe assumption. The only stable property is *which version this repo pins*.

Worse, the failure is quiet in the other direction: a contributor whose PATH ruff is more permissive gets a **clean** hook and a red `Lint (ruff + eslint)` in CI, having done nothing wrong.

## Fix

```json
"*.py": "python3 -m ruff check --fix"
```

Then confirm the same class does not exist elsewhere in the hook path — `eslint --fix` in the same config resolves from `node_modules/.bin`, which is already version-controlled by `package-lock.json`, so it is not the same problem.

## Add the mechanism, not just the fix

Per the ratchet rule added in #1981, this class has now been reported twice — #1676 (in `ci-local.sh`) and here (in `lint-staged`) — and it is mechanically checkable. A check belongs with the fix, along the lines of:

> No hook or lint-staged configuration invokes a version-sensitive linter by bare binary name when the repo pins that tool.

Candidate homes: a `tests/repo/` guard (this is repo tooling, not shipped plugin content, so it does not belong in `repo_invariants.py`'s shipped `CHECKS`). Scope it to the tools this repo actually pins — `ruff` (`requirements-dev.txt`) and `shellcheck` (`SHELLCHECK_VERSION` in `ci-local.sh`) — so a future `"*.sh": "shellcheck"` entry is caught the same way.

## Related, separate

One pre-existing E402 remains tree-wide under 0.15.8 only, in `plugins/dev-team/skills/mutation-testing/scripts/stryker_shard_setup.py`. It is invisible to CI (pinned 0.16.1 calls it clean) and will block any commit touching that file on a machine with an older PATH ruff. It disappears as a practical problem once lint-staged uses the pinned module; worth confirming rather than assuming.

## Correction to the record

An earlier note on [PR #2026](https://github.com/bdfinst/agentic-dev-team/pull/2026) described ruff as "not version-pinned," contrasting it with shellcheck. **That was wrong** — ruff is pinned exactly in `requirements-dev.txt` and has been since #1676. The defect is narrower and more specific: one invocation site bypasses the pin.

### LINKED ISSUE #2033: feat(test-improve): mid-phase mutation-yield steering check at batch boundaries
## Problem

Phase 5 has a mid-phase steering check for coverage and none for mutation.

**Coverage (#1790).** After every Story, `/test-improve` runs `coverage_delta_steering.py` over `coverage-history.json`. Three or more consecutive Stories moving line coverage less than the minimum exits 3 (`flat_streak`) and surfaces `[t] re-check Phase-1 targeting / [c] continue` **mid-phase**. The reference file states plainly why the check exists: *"a run once spent its entire Phase-5 budget on an already-covered layer because this signal was only read at the end."* It also distinguishes `flat_streak_forming` (a watch signal, echoed) from `insufficient_history` (too few Stories to judge) rather than collapsing both into "fine".

**Mutation.** Nothing equivalent. `mutation-kill` is dispatched once per module batch (#1963) with `--max-rounds 3`, and a batch that yields near-zero net kills does not inform the next batch. The lane runs to the end of the Story set regardless of what the earlier batches showed. That is exactly the failure #1790 was written to stop, in the more expensive of the two lanes — `mutation-kill` is 34.4% of thread messages in the 11–25 Aug corpus against 5.1% for every review lens combined.

## Proposed change

Port the #1790 mechanism, not a new one:

1. **Persist per-batch mutation yield.** Append each batch's starting survivor count, ending survivor count, honest score before/after, and rounds spent to `.dev-team-reports/test-improve/<slug>/data/mutation-history.json` — the same git-tracked `data/` sibling as `coverage-history.json`, written atomically the way the baselines are.
2. **A steering script over it,** modeled on `coverage_delta_steering.py` and sharing its status vocabulary and exit-code contract: `ok` / `flat_streak_forming` / `insufficient_history` at exit 0, `flat_streak` at exit 3, unreadable history at exit 2 (**never** read as `ok` — the same trap #1790 calls out).
3. **Surface it at the batch boundary,** where per-Story would be meaningless because mutation-kill now runs per batch. `[t]` re-reads `coverage-gap-ranking.json` and re-orders the remaining batches; `[c]` continues, recording `mutation_flat_streak: <n> batches` in `phase-5.md` so Phase 8 and the Phase-9 report read it from a durable record rather than re-deriving it. Non-interactive runs record and continue — record-and-continue, never a silent pass.

## Interaction with the sibling slices

This composes with the two stop-policy slices filed alongside it but does not depend on either. Those decide when **one file** stops; this decides when the **lane** stops being worth continuing. Both are deterministic scripts over persisted artifacts, per `CLAUDE.md`'s standing rule.

## Blocker to confirm first

`coverage-baseline` and `coverage-delta` appear **nowhere** in the corpus's `skills_invoked` counts, despite Phase 2 and Phase 5 step 3 mandating them. Either they are not instrumented as skill invocations, or the per-Story delta did not fire — and if it is the latter, `coverage_delta_steering.py` had no history to read for the whole run and #1790's check was inert. Resolve which it is before building a second steering check on the same persistence pattern; a mechanism that cannot observe its input is the "gate that cannot fail" case `CLAUDE.md` already warns about.

## Acceptance

- [ ] Determine whether Phase-2/Phase-5 coverage workers actually fired in the observed runs, and fix the instrumentation or the invocation before this slice lands.
- [ ] `mutation-history.json` is written atomically per batch, in the git-tracked `data/` dir.
- [ ] The steering script shares `coverage_delta_steering.py`'s status vocabulary and exit-code contract; unit tests pin `flat_streak`, `flat_streak_forming`, `insufficient_history`, and the unreadable-history exit-2 case.
- [ ] The `[t/c]` prompt fires at a batch boundary mid-phase, never deferred to Phase 9.
- [ ] Non-interactive runs record `mutation_flat_streak` in `phase-5.md` and continue.

Part of #1999

bdfinst:
## Summary

Follow-on to the merged [#2052](https://github.com/bdfinst/agentic-dev-team/pull/2052), continuing the same priority sequence.

**Real bugs (cluster 2)**

- **#1953** — `stryker_shard_pipeline._git` declared `check` keyword-only while the `GitRunner` alias, all three call sites, and every test fake pass it positionally. The shipped default was the one implementation that didn't satisfy its own alias, so **every real invocation raised `TypeError`** — `main()` calls `run_all()` with no `git_run`, so that's all of them. Invisible to the suite because every existing test injects a fake; the new regression tests call the real default.
- **#2027** — `lint-staged` ran the bare `ruff` binary, bypassing #1676's pinned-module rule on the path that runs *every commit*. Measured in this container: pinned 0.16.1 clean, PATH 0.15.8 reporting E402 — and the direction is inverted from #1676's case, so only "which version this repo pins" is stable. Converted to a `tests/repo/` check per the CLAUDE.md ratchet.
- **#1950** — `is_gateway_class_error` missed underscore/camel-joined status keys, `HTTP/1.1 502` (version digits inside the anchor gap), and status **529** (Anthropic's own overload code). Also made the classifier read stdout, which it was structurally blind to.
- **#1952** — `_format_downgrade_message` inferred *why* generation exhausted from a model-name proxy instead of the reason the caller computed and discarded. `DowngradeEvent` now carries `already_downgraded`; the proxy constant is removed.

**Mutation-kill economics (cluster 1 remainder)**

- **#2033** — mid-phase mutation-yield steering at batch boundaries, porting #1790's mechanism (same status vocabulary, same exit-code contract, asserted equal by test) into the lane that is 34.4% of thread messages.
- **#2031** — *partial.* Lands the deterministic core (`survivor_lines`, `confirm_scope_is_complete`, `splice_confirm_over_prior`, `make_confirm_scoped_config`) but **deliberately does not wire it into the loop**.

## Test Plan

- [x] Full pre-push suite green — 9,529 passed, 49 skipped
- [x] `ruff check` clean; `check_md_references.py` clean
- [x] Every fix's tests confirmed **red** against the pre-change code before merge
- [x] #1950 verified against 13 cases including the false-positive guards #1938 anchored for (`request id 502391`, `error after 5024 ms`, `barcode 502391`)
- [x] `HeadlessCallFailed.__str__` asserted unchanged, so no consumer sees new text from the added `stdout` field
- [x] Two repo invariants caught this work and were satisfied properly, not worked around: the Python floor slice (`mutation_yield_steering.py` classified into `SLICE_EXCLUSIONS` with the same reason as its coverage sibling) and the repo-root resolution check
- [ ] #2031 is **not** closed — see below

### Why #2031 is only partial

Its core mechanism depends on Stryker's `mutate` **line-span syntax**, which I could not exercise: there is no dotnet or Stryker in this container, and the repo documents `mutate` only at file-glob granularity. Per CLAUDE.md's own rule, shipping it live would be a guess that looks like a result — and the failure is silent in the confident direction: an unrecognized span matches nothing, the confirm run reports zero mutants, and `_score_round`'s `total_mutants == 0` guard (#1606) reads that as *"NOT convergence"* and stops the file. A wrong guess would truncate every file's loop while looking like a saving.

A test asserts `make_confirm_scoped_config` does not appear in `_score_round`, so wiring it in without recording that verification fails the gate.

### Note on the local gate

`chk_semgrep_fixtures` remains red on this branch **and on `origin/main`** — semgrep crashes on import in this container (`pyo3_runtime.PanicException`, a debian-`cryptography`/pip-`PyJWT` conflict), so every rule reports NO-FIRE because semgrep never runs. Nothing here touches `plugins/security-assessment/`.

Closes #1953
Closes #2027
Closes #1950
Closes #1952
Closes #2033
Part of #2031
Part of #1999

https://claude.ai/code/session_01AY4SPke91uZSVZ3YxdumBF

---
_Generated by [Claude Code](https://claude.ai/code/session_01AY4SPke91uZSVZ3YxdumBF)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,6 +66,47 @@ The local gates (`scripts/ci-local.sh`, run by the `pre-push` hook) need these t
 
 **The pre-push pytest gate spans more than `plugins/dev-team/tests`.** `ci-local.sh`'s unit-test step runs pytest over `plugins/dev-team/tests tests/repo tests/agents tests/commands tests/docs tests/knowledge tests/stack_aware tests/skills tests/scripts tests/hooks` (with `-n auto --dist loadgroup`; two csharp-stryker files `--ignore`d). Editing agent/skill markdown can break repo-level content-guards that live **outside** `plugins/dev-team/tests` — e.g. `tests/skills/test_code_review_frontend_dispatch.py` asserts specific agent names appear verbatim in `code-review/SKILL.md`, and `tests/repo` runs `check_md_references.py` (a backticked cross-skill path must be file-relative, e.g. `../code-review/SKILL.md`). `tests/hooks/` (repo-root, distinct from `plugins/dev-team/tests/hooks/`) joined this list in #1475 after two of its tests — pinning `hooks.json`'s dispatch-matcher registration — went silently red for a full ADR migration (ADR 0026) because this directory wasn't in the gate; the stryker/pitest/mutmut *adapter* tests that actually shell out to those tools live in `plugins/dev-team/tests/hooks/` (already covered by `plugins/dev-team/tests` above) — this directory only hosts the static fixture files those adapter tests read (`tests/hooks/fixtures/`, `tests/hooks/fake-bin/`), so `tests/hooks/` itself is fast and portable. Before pushing plugin-content changes, run that **full dir list** — not just the plugin subdir — plus `python3 scripts/check_md_references.py` and `python3 plugins/dev-team/hooks/lib/build_knowledge_index.py`, or the `pre-push` hook / CI will catch what a plugin-only run missed.
 
+### The inner loop — three speeds, not two
+
+Before #2002/#2005 the loop had two speeds: run one file, or run the whole
+9,469-test directory list. "What does this change affect?" was answered by a
+guess, and per the deterministic-tools rule that is a mechanical question a
+program should compute.
+
+1. **`--lf` / `--ff`** — free, built into pytest, no map and no dependency.
+   `--lf` re-runs only last run's failures; `--ff` runs them first, then the
+   rest. This is the right tool for the tight red→green loop and costs nothing
+   to adopt.
+2. **Impact selection** — `scripts/impact_tests.py` answers "given that
+   something changed, which tests reach it" from a per-test coverage map:
+
+   ```bash
+   python3 scripts/impact_tests.py build --out .cache/impact-map.json -- \
+     plugins/dev-team/tests tests/repo tests/agents tests/commands tests/docs \
+     tests/knowledge tests/stack_aware tests/skills tests/scripts tests/hooks
+   python3 scripts/impact_tests.py select --map .cache/impact-map.json \
+     --changed-from-git | xargs python3 -m pytest -q
+   ```
+
+   Built on `pytest-cov`'s `--cov-context=test` — already a dev dependency —
+   rather than adding `pytest-testmon`. Measured on this checkout: a
+   `hooks/telemetry.py` edit selects **28 tests in 0.6s** against 9,493 in
+   ~80s. Building the map costs one full instrumented run (~4 min).
+
+   **It never narrows on a guess.** `select` exits **2** — meaning *run the
+   full suite* — when the map is missing or malformed, when a changed file is
+   absent from the map (a new or uncovered file has unknown reach), or when a
+   changed file is itself a test. Treat any non-zero exit as the full suite,
+   never as an empty selection. The map is a snapshot: rebuild it after adding
+   tests or source files.
+3. **The full directory list** — still what `pre-push` and CI run, and still
+   the answer whenever selection refuses.
+
+`scripts/ci_tree_cache.py` (#2002) sits underneath all three and answers a
+different question — "has anything changed at all" — skipping `chk_hook_units`
+outright when the working tree is byte-identical to the one it last passed on.
+`CI_LOCAL_NO_TREE_CACHE=1` forces a run.
+
 ### Worktrees — run `npm ci` first
 
 Run `npm ci` as the first step in any new worktree, before committing. An unprovisioned worktree has no `.husky/_`/`node_modules`, so git hooks silently don't run; `scripts/ci-local.sh` and CI backstop what the hooks would have caught.
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,mjs,cjs,ts,tsx}": "eslint --fix",
-    "*.py": "ruff check --fix"
+    "*.py": "python3 -m ruff check --fix"
   },
   "devDependencies": {
     "@commitlint/cli": "^21.0.2",
--- a/plugins/dev-team/docs/test-improve.md
+++ b/plugins/dev-team/docs/test-improve.md
@@ -84,9 +84,14 @@ is `none`, +1 once Phase 6 enters Phase 7) rather than a hardcoded 9 or 10.
   (no-refactor) → `/coverage-delta --workflow test-improve --story <id>` →
   `scripts/coverage_delta_steering.py` (three consecutive near-zero-delta
   Stories exit 3 and prompt `[t] re-check Phase-1 targeting / [c] continue`
-  mid-phase, #1790) → `mutation-kill` agent (`--file <story-file>
-  --max-rounds 3`, `[c/r/w/q]` on
-  residuals). End-of-phase review loop runs `/test-design --since` and
+  mid-phase, #1790) → `mutation-kill` agent, dispatched once per module batch
+  (`--file <every story file in the batch> --max-rounds 3
+  --target-honest-score <Phase-0 mutation target>`, #2030; `[c/r/w/q]` on
+  residuals) → `scripts/mutation_yield_steering.py` at the batch boundary
+  (two consecutive batches killing fewer than the minimum net survivors exit
+  3 and prompt the same `[t] re-check Phase-1 targeting / [c] continue`,
+  #2033 — the #1790 mechanism ported to the more expensive lane, sharing its
+  status vocabulary and exit-code contract). End-of-phase review loop runs `/test-design --since` and
   `/code-review --since` in parallel, `/apply-fixes` then re-run, cap 2
   iterations, `[r/w/q]` escalation. Evidence in `phase-5-review.json`.
 - **Phase 6 — Refactor decision prompt.** `[y] enter Phase 7 / [b] backlog
--- a/plugins/dev-team/scripts/mutation_yield_steering.py
+++ b/plugins/dev-team/scripts/mutation_yield_steering.py
@@ -1,0 +1,345 @@
+#!/usr/bin/env python3
+"""mutation_yield_steering.py — turn per-batch mutation yield into a live
+Phase-5 steering signal (issue #2033).
+
+Why this exists. Phase 5 has a mid-phase steering check for coverage (#1790,
+``coverage_delta_steering.py``) and none for mutation. `mutation-kill` is
+dispatched once per module batch (#1963) with ``--max-rounds 3``, and a batch
+that yields near-zero net kills does not inform the next batch — the lane runs
+to the end of the Story set regardless of what the earlier batches showed.
+That is exactly the failure #1790 was written to stop, in the more expensive of
+the two lanes: `mutation-kill` is 34.4% of thread messages in the 11-25 Aug
+2026 corpus against 5.1% for every review lens combined.
+
+This is a **port of #1790's mechanism, not a new one**. It shares that
+script's status vocabulary and exit-code contract deliberately, so an operator
+reading one already knows how to read the other, and so `/test-improve` can
+branch on both identically.
+
+The unit is the **batch**, not the Story. `mutation-kill` now runs per batch,
+so a per-Story mutation check would have no observation to make on most
+Stories — the signal only exists at a batch boundary.
+
+Yield, per batch, is ``starting_survivors - ending_survivors``: the net
+survivors killed. A **negative** yield counts as flat — a batch that ended
+with more survivors than it started is certainly not progress. A batch whose
+yield cannot be measured (either count missing or non-numeric) has yield
+``None`` and **breaks** the streak rather than extending it: absence of
+measurement is not evidence of flatness.
+
+Honest score before/after is reported alongside but does **not** drive the
+verdict. Score movement depends on the size of the file's mutant set, so a
+batch over a large module can kill real survivors while barely moving the
+percentage; keying the gate on it would make the check silently
+module-size-dependent. This mirrors #1790's reason for not keying on branch
+coverage.
+
+Statuses (and exit codes):
+
+- ``flat_streak`` (3) — ``--consecutive`` or more trailing batches each yielded
+  fewer than ``--min-kills``. The caller prompts; this script does not decide
+  policy.
+- ``flat_streak_forming`` (0) — the trailing batch yielded less than the
+  minimum but the streak is still shorter than ``--consecutive``. A distinct
+  status rather than ``ok``, because "not yet a streak" is not "mutation-kill
+  is producing kills".
+- ``insufficient_history`` (0) — fewer measurable batches than
+  ``--consecutive``, or a trailing batch whose yield could not be measured at
+  all, so no streak verdict is possible yet.
+- ``ok`` (0) — the trailing batch yielded at least the minimum.
+- exit ``2`` — the history file is missing, unreadable, not JSON, not a JSON
+  array, or carries a non-record element; also an out-of-range
+  ``--consecutive``/``--min-kills``. **Never** reported as an all-clear — the
+  same trap #1790 calls out.
+
+Stdlib-only (ADR 0014/0015), Python 3.10+ floor (ADR 0031).
+
+Usage:
+    python3 mutation_yield_steering.py \\
+      --history .dev-team-reports/test-improve/<slug>/data/mutation-history.json
+    python3 mutation_yield_steering.py --history <path> \\
+      --min-kills 1 --consecutive 2 --json
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+DEFAULT_MIN_KILLS = 1
+DEFAULT_CONSECUTIVE = 2
+
+STATUS_FLAT = "flat_streak"
+STATUS_FORMING = "flat_streak_forming"
+STATUS_OK = "ok"
+STATUS_INSUFFICIENT = "insufficient_history"
+
+# Score movements are rounded only to kill binary-float artifacts
+# (72.4 - 72.0 = 0.40000000000000568). Yield itself is an integer count and is
+# never rounded. Score never decides the verdict — see the module docstring.
+_SCORE_PRECISION = 10
+
+
+class HistoryError(ValueError):
+    """The mutation history could not be read or is not a record array."""
+
+
+def load_history(path: Path) -> list[dict]:
+    """Read ``mutation-history.json``, raising HistoryError on anything that is
+    not a readable JSON array of batch-record objects."""
+    if not path.is_file():
+        raise HistoryError(f"mutation history not found: {path}")
+    try:
+        payload = json.loads(path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError) as exc:
+        raise HistoryError(f"{path}: unreadable mutation history ({exc})") from exc
+    if not isinstance(payload, list):
+        raise HistoryError(
+            f"{path}: mutation history must be a JSON array of batch records, "
+            f"got {type(payload).__name__}"
+        )
+    # A non-dict element means the file is corrupt or half-rewritten. Dropping
+    # it silently would leave a corrupt history indistinguishable from a phase
+    # that simply hasn't closed enough batches yet — a clean verdict from
+    # unusable evidence, which is the one thing this gate must never produce.
+    for index, entry in enumerate(payload):
+        if not isinstance(entry, dict):
+            raise HistoryError(
+                f"{path}: mutation history entry at index {index} is a "
+                f"{type(entry).__name__}, not a batch record"
+            )
+    return list(payload)
+
+
+def batch_records(history: list[dict]) -> list[dict]:
+    """Keep only batch-attributed records — the per-batch Phase-5 dispatches.
+
+    A record with no ``batch`` is not a batch boundary (e.g. a whole-suite
+    Phase-8 measurement written into the same stream) and must not enter the
+    streak. Mirrors #1790's ``story_snapshots``.
+    """
+    return [entry for entry in history if entry.get("batch")]
+
+
+def _as_number(value: object) -> float | None:
+    if isinstance(value, bool):
+        return None
+    if not isinstance(value, (int, float)):
+        return None
+    return float(value)
+
+
+def _yield_of(record: dict) -> int | None:
+    """Net survivors killed by this batch, or ``None`` when unmeasurable."""
+    start = _as_number(record.get("starting_survivors"))
+    end = _as_number(record.get("ending_survivors"))
+    if start is None or end is None:
+        return None
+    return int(start - end)
+
+
+def _score_movement(record: dict) -> float | None:
+    before = _as_number(record.get("honest_score_before"))
+    after = _as_number(record.get("honest_score_after"))
+    if before is None or after is None:
+        return None
+    return round(after - before, _SCORE_PRECISION)
+
+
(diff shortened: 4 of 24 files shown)


bdfinst:
## Session handoff — priority-sequenced open-issue work

Posted here rather than as a file because `.claude/memory/` is gitignored and this session's container is ephemeral.

**Branch:** `claude/open-issues-data-dependency-815jmu` · **Merged:** #2052 (#2001, #2030) · **12 of 30 issues done**

### Sequence and status

| # | Cluster | Issues | Status |
|---|---------|--------|--------|
| 1 | Mutation-kill economics | #2001, #2030, #2031, #2033 | **done** (#2031 partial) |
| 2 | Real bugs | #1953, #2027, #1950, #1952 | **done** |
| 3 | Dev-loop speed | #2002, #2003, #2005 | **done** |
| 4 | Sensor-integrity fixes | #1998, #2036, #2018, #2016 | not started |
| 5 | Debt | #1708, #1716, #1717, #1723 | not started |
| 6 | Modeling decisions | #1948, #1951, #1955, #1956 | not started |
| 7 | Unified session pipeline | #2040 + #2041–#2050 | not started |

### Start here

Cluster 4 is next in sequence *and* the highest leverage in the backlog — those four are what the largest cluster of other issues divides by.

- **#1998** — 18.2% of review-agent outputs fail the JSON contract and are silently dropped, varying ~8× by agent. Explicitly blocks #2007, #1982, #1980. The issue asks to **instrument before fixing**.
- **#2036** — `_BYPASS_RE`/`_COMMIT_RE` (`session_extract.py:455-458`) are unanchored, so `git commit -m x && grep -n TODO src/` counts as a bypass. Ships standalone. Keep `hooks/telemetry.py` in agreement — the comment at `session_extract.py:76-78` says the two mirror each other.
- **#2018** — 30-day evidence TTL + `plugin_version` stamped at *extraction* time.
- **#2016** — `rollup()` ingests peer keys ungated. `_read_synced_records()` is the natural choke point.

### Environment facts that cost real time

1. **Never run `ruff format`.** `chk_ruff` is `ruff check` only. Running format rewrote **70 untouched files** and had to be reverted individually.
2. **semgrep is broken in this container**, so `chk_semgrep_fixtures` is red here **and on `origin/main`** — it crashes on import (`pyo3_runtime.PanicException`, debian-`cryptography`/pip-`PyJWT` conflict), so ~20 rules report NO-FIRE because the tool never runs. `pip` can't repair it (PyPI unreachable over TLS despite being in the proxy's `noProxy`). **Pushes therefore need `--no-verify`**, with every other gate run manually first.
3. **The context-ceiling hook blocks `Skill` invocations** above 150k (ADR 0037) — run `bash scripts/dev-setup.sh` directly instead of `/setup`. Note its verify step doesn't check pytest, the most-used dev dep.
4. **`.coveragerc` sets `branch = True`**, so coverage writes to the `arc` table, not `line_bits` — a `line_bits`-only query silently returns empty. Its `source` also omits `plugins/dev-team/scripts`.
5. **Three repo invariants catch new files** — the Python floor slice, repo-root resolution (`from _repo_root import REPO_ROOT`), and anchor citations in agent files. Satisfy them; don't work around.

### Deliberately unfinished

**#2031 is `Part of`, not `Closes`, and must stay that way.** The deterministic core shipped but is **not wired into the loop**: its mechanism needs Stryker's `mutate` line-span syntax, unverifiable here (no dotnet/Stryker, and the repo documents `mutate` only at file-glob granularity). A wrong guess fails silently in the confident direction — an unrecognized span matches nothing → zero mutants → `_score_round`'s `total_mutants == 0` guard (#1606) reads that as *"NOT convergence"* and stops the file, truncating every loop while looking like a saving. A test asserts `make_confirm_scoped_config` never appears in `_score_round`, so wiring it in without recording the verification fails the gate.

**#2030 AC4** ("Phase 8's verdict unchanged, verified on one real module batch") needs a real dotnet/mutmut run. Not covered.

### Two specs that turned out to be wrong

Worth knowing the issues aren't always right:

- **#2003** proposed seeding the inert-path set with `.gitignore`. The guard test proved the suite reads the root `.gitignore` in three places — a false skip there would have been silent and CI-only. Shipped as `LICENSE .gitattributes` only; the proposed `.editorconfig` and `CODEOWNERS` don't exist in this repo.
- **#2005** named `pytest-testmon`. `pytest-cov`'s `--cov-context=test` — already a dependency — gives the same per-test map, so no dependency was added.

### Conventions used

One commit per issue, conventional title, `Closes`/`Part of` in the body. **Every fix's tests were confirmed RED against pre-change code** before merge — that caught three tests that would otherwise have passed vacuously. Full gate before every push (~9,586 passing at handoff).

---

bdfinst:
### Correction to the handoff above

Two lines in the handoff went stale within minutes of posting, and since it's written for another session to act on, they're worth fixing rather than leaving.

**1. The inert set is `LICENSE` only — not `LICENSE .gitattributes`.**

CI went red on `63dc2b7`/`d9d1b4d`/`7843537` with one failing test, mine:

```
test_every_inert_path_actually_exists_in_the_repo
AssertionError: inert paths that do not exist in this repo: ['.gitattributes']
```

This repo tracks no `.gitattributes`. The one on the dev container is a **gitignored artifact graphify's hook install writes** (`graphify-out/graph.json merge=graphify`), created when `dev-setup.sh` ran. My seed probe used `exists()` on the filesystem, saw the artifact, and concluded it was a repo file — and so did the guard test written to catch exactly that class of error. Fixed in `eea2b9b`: `.gitattributes` dropped, and the guard now asks `git ls-files` instead of the filesystem.

**2. Add a sixth environment fact — and it's the one that actually bit.**

> **`dev-setup.sh` mutates the working tree.** graphify's hook install writes a gitignored `.gitattributes`; other tooling may write more. A check that asks the *filesystem* whether something is a repo file will therefore answer differently on a developer machine and on a runner. Ask **git**, not the disk.

That's the platform-divergence class `CLAUDE.md` documents for `mktemp` and the interpreter gates, on a new axis — an untracked file rather than an OS or a version. It's now the second time this container has diverged from CI (semgrep being broken here was the first), so treat local green as necessary, not sufficient.

Verified by reproducing the divergence rather than reasoning about it: with the local `.gitattributes` moved aside (simulating the runner) the suite is green, and green again with it restored. The old shape passed in one state and failed in the other.

Everything else in the handoff stands.

---

Issue #2159, opened 2026-09-15
feat(specs): epic — close DeFOSPAM spec-review gaps


bdfinst:
Umbrella for closing the four highest-priority gaps found in `.dev-team-reports/competitive-analysis-2026-09-15.md`, comparing our spec-review process against [AgenticTesting/OpenRequirementsAI](https://github.com/AgenticTesting/OpenRequirementsAI) (the DeFOSPAM seven-analyst requirements-validation skill).

## Baseline (from the analysis)

The comparison found that coverage is mostly **not** the problem — `plan-review-acceptance` already owns binary verifiability (weasel-word detection), boundary completeness, error-path coverage, negative testing, and state transitions, all at `blocker` severity. What it does not own is **placement and intake**:

- `/specs` is a collaborative *authoring* loop ("Human drafts based on current understanding"). There is no path that takes a requirements document someone else wrote and validates it. A stakeholder PRD, RFP, or Jira epic reaches `/plan` unexamined.
- Every per-criterion check runs against criteria **we** authored. Checking our own restatement cannot catch an ambiguity that was resolved — possibly wrongly — during restatement.
- Nothing sweeps for *absent* entities or concern categories. A spec that never mentions deletion produces no criterion for `plan-review-acceptance` to find incomplete; the omission is invisible to every gate we run.
- No spec-time glossary exists, so an undefined domain term reaches `/plan` as an ordinary word. `ubiquitous-language` only engages once terms are already in use.

## Slices

| # | Slice | Priority | Why |
|---|-------|----------|-----|
| 1 | Validate an externally-authored requirements document | High | The only true workflow-level hole — changes what the plugin can *do* |
| 2 | Completeness sweep (CRUD + cross-cutting concerns) | High | Catches whole-entity omissions that are structurally invisible to per-criterion checks |
| 3 | Glossary artifact with per-term verification status | Medium | Quick win with an obvious downstream consumer in `ubiquitous-language` |
| 4 | Predictability check inside the ambiguity protocol | Medium | Attacks the Ambiguity Resolution Protocol's own stated failure mode |

Sub-issues are attached below in priority order. PRs addressing a slice carry `Closes #<sub-issue>` and `Part of #<this epic>`; `epic-auto-close` (#987) closes this epic when the last sub-issue closes.

## Constraints carried across every slice

[...]

PR #2180, opened 2026-09-16
feat(specs): close the DeFOSPAM spec-review placement gaps

### LINKED ISSUE #2160: feat(specs): validate an externally-authored requirements document
## Intent Description

`/specs` today is a collaborative authoring loop — it assumes the human is drafting the requirement with us, right now ("Human drafts based on current understanding"). There is no entry point for a requirements document that already exists and that we did not help write: a stakeholder PRD, a vendor RFP, a Jira epic, a customer's written brief.

Today such a document reaches `/plan` unexamined. Whoever is driving reads it, forms a mental model, and starts authoring criteria from that model — so the first artifact any gate sees is already our *restatement*. Every ambiguity the source contained has by then been silently resolved, correctly or not.

This slice adds a validate mode to `/specs` that takes the document as input and runs the existing critique categories against the **source text**, producing the same three artifacts and the same Ambiguity Log, with each extracted criterion citing the source passage it came from. The goal is that a third-party document hits the same `requires-stakeholder-input` hard block an in-house draft does.

Part of epic #2159.

## Architecture Specification

- **Entry point**: `skills/specs/SKILL.md`, a new mode selected at Step 0. Mode is chosen by inspecting the argument, not by a flag: an argument that resolves to a readable file path or a GitHub issue URL selects validate mode; anything else is a prose feature description and selects today's authoring mode. Existing invocations are unaffected.
- **Supported inputs**: whatever the `Read` tool handles natively — `.md`, `.txt`, `.pdf`. `.docx` is explicitly **not** supported; the plugin ships stdlib-only Python (ADR 0014/0015) and no stdlib path parses it. On an unsupported extension, say so and name the conversion the human should do. Do not attempt a partial parse.
- **Extraction**: read the document, then populate the three existing artifacts (Intent Description, Architecture Specification, Acceptance Criteria) from its content. Each extracted acceptance criterion carries a citation — a section heading, a line reference, or a short verbatim quote — identifying where in the source it came from. A criterion with no citable source passage is an inference, not an extraction, and is logged as such.
- **Critique**: the existing four critique categories (gap, ambiguity, conflict, scope violation) run unchanged, against the source text rather than against a co-authored draft.
- **Routing**: findings enter the existing Ambiguity Resolution Protocol and Ambiguity Log. No new classification vocabulary, no new gate, no confidence score.
- **Persistence**: unchanged — the existing file / GitHub-issue branch in "Classify where to persist" applies as-is.
- **Token budget**: the SKILL.md delta is a short mode-selection step plus the citation rule. Extraction heuristics, if they grow beyond a few lines, go in `knowledge/` and load on demand (epic #2159 constraint).

## Acceptance Criteria

- Invoking `/specs` with a path to a readable `.md`, `.txt`, or `.pdf` file enters validate mode and produces the three artifacts plus an Ambiguity Log, without the human having drafted anything first.
- Invoking `/specs` with a GitHub issue URL enters validate mode against that issue's body.
- Invoking `/specs` with a prose feature description still enters today's authoring mode, with behavior byte-for-byte unchanged.
- Every extracted acceptance criterion carries a citation to the source passage it was derived from.
- A criterion with no citable source passage is recorded in the Ambiguity Log as an inference, never presented as if the source stated it.
- An unsupported input format (`.docx` and anything else `Read` cannot handle) is refused with a named reason and a suggested conversion — never partially parsed.
- A `requires-stakeholder-input` finding from a validated third-party document blocks before acceptance criteria are finalized, exactly as it does in authoring mode.
- The Consistency Gate runs and must pass before the mode completes.
- No new gate, severity scheme, or confidence score is introduced.

## Ambiguity Log

| Decision | Classification | Resolved By | Rationale / Answer |
|----------|---------------|-------------|-------------------|
| How validate mode is selected | `inferable` | inference | Argument shape is unambiguous — a resolvable path or issue URL vs. prose. Avoids a flag, and avoids breaking every existing `/specs "<description>"` invocation. |
| Supported input formats | `inferable` | inference | Bounded by what `Read` handles natively and by the stdlib-only constraint (ADR 0014/0015). DeFOSPAM supports `.docx`; we cannot without a dependency, so refusing loudly beats parsing badly. |
| Whether validate mode still blocks on the human | `inferable` | inference | The protocol's whole value is "resolving ambiguity with a human before build begins" (skill's own Positioning section). A third-party document has *more* unresolved ambiguity, not less — weakening the block here would invert the intent. |
| Criterion-to-source citation | `inferable` | inference | Without it, extraction and inference are indistinguishable in the output, which is the exact failure mode this slice exists to prevent. Mirrors DeFOSPAM's traceability property (every finding links to its source excerpt). |
| Reuse of existing persistence branching | `inferable` | inference | The origin/marker classification in "Classify where to persist" is orthogonal to how the spec was produced. No reason to fork it. |
| **Does validate mode auto-trigger `/plan`?** | `requires-stakeholder-input` | **unresolved — blocks final criteria** | Today's auto-trigger has an explicit "do not ask first" contract, justified because the human just co-authored and approved. Validating an RFP or a competitor's spec carries no build commitment, so auto-planning could be actively wrong. **Proposed default**: validate mode does *not* auto-trigger `/plan`; it prints the persisted location and offers `/plan` as an explicit next step. Needs confirmation — if you want symmetry with authoring mode instead, say so and the criteria change. |

## Consistency Gate
- [x] Intent is unambiguous
- [x] Every behavior/goal maps to an acceptance criterion
- [x] Architecture constrains without over-engineering
- [x] Terminology consistent across artifacts
- [x] No contradictions between artifacts
- [ ] Every gap/ambiguity finding is logged — inferable with rationale or resolved by human — **one `requires-stakeholder-input` item is unresolved (see Ambiguity Log); criteria are not final until it is answered**

### LINKED ISSUE #2161: feat(specs): completeness sweep for CRUD and cross-cutting concerns
## Intent Description

`plan-review-acceptance` checks whether each acceptance criterion the spec contains is complete — boundaries at zero/one/many, an error path per happy path, negative criteria, illegal state transitions. That is a strong per-criterion check, and this slice must not duplicate any of it.

It answers a different question from the one that matters here. A spec that never mentions deletion at all produces **no criterion** for `plan-review-acceptance` to find incomplete. The omission is not a weak criterion; it is the absence of one, and absence is structurally invisible to every gate we currently run. The same holds for authorization, audit, and error handling when the spec simply never raises them.

This slice adds a sweep that enumerates the entities and capabilities the spec names, then checks each against a fixed checklist of operations and cross-cutting concerns — so that what the spec *failed to mention* becomes a finding, not a silence.

Part of epic #2159.

## Architecture Specification

- **Placement**: a new step in `skills/specs/SKILL.md` running after the critique loop and **before** the Cross-Artifact Consistency Gate, so its findings are available to the Ambiguity Resolution Protocol.
- **Checklist location**: `plugins/dev-team/knowledge/completeness-checklist.md`, loaded on demand. The SKILL.md delta is the step and its routing only, not the checklist body (epic #2159 progressive-disclosure constraint).
- **Checklist content** (fixed, shipped):
  - CRUD coverage per named entity — create, read, update, delete.
  - Cross-cutting concerns — authentication, authorization, audit/logging, error handling.
  - Administrative and non-functional surfaces the spec's own domain implies.
- **Not extensible per project.** One shipped checklist. A project-override mechanism is not in scope and should not be built speculatively.
- **Entity enumeration is a model step, not a deterministic one.** Extracting entities from prose is not something a parser answers, so CLAUDE.md's deterministic-tools-first rule is satisfied at the level it can be: the *checklist application* against an enumerated entity is mechanical and fixed, even though the enumeration is inferred. The step must say which entities it enumerated, so a human can see an entity that was missed.
- **Routing**: each unaddressed checklist cell becomes an Ambiguity Log entry, classified `inferable` (with rationale — e.g. "this entity is read-only by design, stated in the architecture notes") or `requires-stakeholder-input`. **The sweep is not itself a gate.** It blocks only through the existing protocol, exactly as every other finding does.
- **Division of labor**: the sweep never evaluates the quality or completeness of a criterion that exists — that is `plan-review-acceptance`'s scope. It only reports cells with no corresponding criterion at all.

## Acceptance Criteria

- The sweep runs after the critique loop and before the Consistency Gate.
- The sweep reports the list of entities and capabilities it enumerated, so a missed entity is visible to the human.
- Each enumerated entity is checked for all four CRUD operations.
- Each spec is checked for the four cross-cutting concerns (authn, authz, audit/logging, error handling).
- Every unaddressed cell produces an Ambiguity Log entry classified `inferable` with rationale or `requires-stakeholder-input`.
- A cell deliberately not applicable is recorded as `inferable` with its rationale, never silently dropped.
- The sweep introduces no new gate, severity scheme, or confidence score — it blocks only via the existing protocol.
- The sweep produces no finding about the quality of a criterion that already exists (no overlap with `plan-review-acceptance`).
- The checklist lives in `knowledge/completeness-checklist.md` and loads on demand; the `skills/specs/SKILL.md` delta does not inline it.

## Ambiguity Log

| Decision | Classification | Resolved By | Rationale / Answer |
|----------|---------------|-------------|-------------------|
| Sweep placement in the `/specs` flow | `inferable` | inference | Must precede the Consistency Gate for its findings to reach the Ambiguity Resolution Protocol; must follow the critique loop so it sweeps the refined artifacts, not a first draft. |
| Fixed checklist vs. per-project extensible | `inferable` | inference | Ship one fixed list. Per-project override is a hypothetical requirement with no current caller; building it now is speculative design. |
| Sweep is a finding source, not a gate | `inferable` | inference | Epic #2159's one-triage-vocabulary constraint. A second gate would give two ways for a spec to be blocked and two vocabularies to reconcile. |
| Entity enumeration by model, not parser | `inferable` | inference | No parser extracts domain entities from prose. CLAUDE.md's rule is "prefer, in order: run the real thing, a deterministic script, a model, only for what's left" — enumeration is genuinely in the last bucket. Mitigated by requiring the enumeration be shown. |
| Checklist scope (CRUD + 4 cross-cutting concerns) | `inferable` | inference | Mirrors Milarna's documented sweep (CRUD, admin/non-functional, auth/audit/error-handling) — the capability this slice is closing. Starting narrower would not close the gap; starting wider invents categories with no evidence behind them. |
| No overlap with `plan-review-acceptance` | `inferable` | inference | Epic #2159 constraint. The two answer different questions: "is this criterion complete" vs. "is there a criterion here at all." Stated explicitly so the boundary survives future edits. |

## Consistency Gate
- [x] Intent is unambiguous
- [x] Every behavior/goal maps to an acceptance criterion
- [x] Architecture constrains without over-engineering
- [x] Terminology consistent across artifacts
- [x] No contradictions between artifacts
- [x] Every gap/ambiguity finding is logged — inferable with rationale or resolved by human

### LINKED ISSUE #2162: feat(specs): glossary artifact with per-term verification status
## Intent Description

`ubiquitous-language` enforces terminology consistency across code and docs, but only once terms are already in use. Nothing captures, at spec time, which domain terms are still undefined — so a term nobody has agreed on reaches `/plan` looking like an ordinary word, and the disagreement surfaces later as a naming argument or, worse, as two components that mean different things by "order".

This slice adds a fourth spec artifact: a glossary that names each domain term, its definition, and whether a human has actually verified that definition. An unverified term becomes a gap finding rather than a silence, and the agreed vocabulary is handed downstream instead of being re-derived.

Part of epic #2159.

## Architecture Specification

- **Artifact**: a `## Glossary` section added to both persistence templates in `skills/specs/SKILL.md` — the `docs/specs/<slug>.md` file template and the GitHub-issue body template. Both must carry it; the two templates already mirror each other and must not diverge.
- **Shape**: a table of term / definition / verification status / source, where status is exactly `verified` or `unverified`. `verified` means a human confirmed the definition during the collaboration loop — not that an agent found a plausible one.
- **Population**: terms are extracted while drafting Intent and Acceptance Criteria, as part of the existing collaboration loop rather than as a separate pass. A term whose definition the agent inferred starts `unverified` and stays that way until the human confirms it.
- **Routing**: an `unverified` term at the end of the loop is a gap finding and enters the Ambiguity Resolution Protocol, classified `inferable` (the definition follows unambiguously from codebase or domain convention) or `requires-stakeholder-input` (two reasonable people would define it differently). **No new gate** — an unverified term blocks only if the protocol classifies it as blocking.
- **Downstream handoff**: the glossary is part of the persisted spec artifact, so `ubiquitous-language` and `domain-review` consume it by reading the spec they already locate. No new index, no new file, no new lookup mechanism.
- **Token budget**: SKILL.md delta is the template section plus the routing sentence. If term-extraction heuristics grow past a few lines, they go in `knowledge/` (epic #2159 constraint).

## Acceptance Criteria

- Both persistence templates — file and GitHub issue — contain a `## Glossary` section, with identical structure.
- Each glossary row carries term, definition, verification status, and source.
- Verification status is exactly `verified` or `unverified`; no third value.
- A definition the agent inferred is recorded `unverified` until a human confirms it.
- Every term still `unverified` at the end of the collaboration loop produces an Ambiguity Log entry classified `inferable` with rationale or `requires-stakeholder-input`.
- An unverified term does not by itself block the Consistency Gate — it blocks only via the existing protocol classification.
- A spec with no domain terms worth defining renders the section per the empty-section convention rather than omitting it silently.
- The glossary is persisted inside the spec artifact; no separate glossary file or index is created.
- No new gate, severity scheme, or confidence score is introduced.

## Ambiguity Log

| Decision | Classification | Resolved By | Rationale / Answer |
|----------|---------------|-------------|-------------------|
| Glossary as a spec artifact vs. a standalone file | `inferable` | inference | A standalone file needs its own discovery, staleness, and lifecycle story. Inside the spec, existing consumers (`ubiquitous-language`, `domain-review`, `spec-compliance-review`) find it via the spec they already locate. |
| Two-valued verification status | `inferable` | inference | Mirrors Dorothy's `verified`/`unverified`. A third state ("proposed", "disputed") has no consumer and would need its own routing rules. |
| `verified` requires a human, not an agent | `inferable` | inference | An agent-confirmed definition is exactly the "looks thorough while encoding a happy-path assumption" failure the Ambiguity Resolution Protocol exists to prevent. If an agent could self-verify, the status would carry no information. |
| Unverified terms route through the existing protocol rather than blocking directly | `inferable` | inference | Epic #2159 one-triage-vocabulary constraint. Blocking on every unverified term would make specs painful enough that people route around `/specs` entirely. |
| Extraction during the loop vs. a separate pass | `inferable` | inference | Terms surface naturally while drafting intent and criteria; a separate pass would re-read the same artifacts to find the same words. |
| Both templates must carry the section | `inferable` | inference | The file and issue templates are already mirrors. Divergence would mean a spec's completeness depends on which persistence branch fired — a silent behavioral difference. |

## Consistency Gate
- [x] Intent is unambiguous
- [x] Every behavior/goal maps to an acceptance criterion
- [x] Architecture constrains without over-engineering
- [x] Terminology consistent across artifacts
- [x] No contradictions between artifacts
- [x] Every gap/ambiguity finding is logged — inferable with rationale or resolved by human

### LINKED ISSUE #2163: feat(specs): predictability check inside the ambiguity protocol
## Intent Description

The Ambiguity Resolution Protocol names its own failure mode plainly: spec synthesis tends to produce "decisions that look thorough while encoding the same happy-path assumptions a direct implementation would make silently." The `inferable` classification is where that failure lands — an assumption gets waved through because it reads as natural, and naturalness is explicitly *not* the test.

`plan-review-acceptance` already applies the right test — binary verifiability, "can two people independently check this criterion and agree on pass/fail" — at `blocker` severity. But it runs one stage later, against criteria **we** authored. Checking our own restatement for predictability cannot catch a source requirement that was unpredictable *before* we normalised it: by then the ambiguity has already been resolved, and the criterion reads cleanly precisely because someone picked an interpretation.

This slice adds a small, concrete test inside the protocol that makes an assumption falsifiable at the point where it is being classified: for each criterion, name a plausible alternative outcome and ask whether the source rules it out.

Part of epic #2159.

## Architecture Specification

- **Placement**: inside the Ambiguity Resolution Protocol in `skills/specs/SKILL.md`, between Step A (attempt inference) and Step B (classify the finding). It is a test applied *during* classification, not a separate pass over the artifacts.
- **The test**: for each acceptance criterion, generate exactly **one** plausible alternative outcome — a behavior a reasonable developer might implement instead — and check whether the spec text rules it out. If it does not, the criterion is `requires-stakeholder-input`, not `inferable`.
- **One alternative, not several.** The check is a tripwire, not an enumeration. Generating many alternatives per criterion would turn a classification aid into its own analysis phase and bloat both the step and the Ambiguity Log.
- **Plausible, not absurd.** DeFOSPAM's Paul floats deliberately absurd alternatives to provoke stakeholder correction. That works in an advisory tool whose findings are suggestions. Here the output flips a blocking classification, so an absurd alternative would manufacture false blocks and train people to wave the check through. The alternative must be one a competent developer could actually ship.
- **Output**: when the check flips a classification, the alternative outcome is recorded in the Ambiguity Log row's rationale — it is the evidence for why the item blocks, and it is what the human answers.
- **Division of labor**: this does not duplicate `plan-review-acceptance`'s binary-verifiability check, which still runs later and still owns weasel-word detection and per-criterion completeness. This slice's addition is *placement* — catching unpredictability in the source before restatement hides it. The SKILL.md text must say so, so a future reader does not fold the two together as redundant.
- **Token budget**: a short step. No `knowledge/` file needed.

## Acceptance Criteria

- The predictability check runs inside the Ambiguity Resolution Protocol, between Step A and Step B.
- For each acceptance criterion, exactly one plausible alternative outcome is generated and tested against the source text.
- A criterion whose alternative outcome the source does not rule out is classified `requires-stakeholder-input`, never `inferable`.
- The generated alternative is recorded in the Ambiguity Log row's rationale whenever it changes a classification.
- The alternative is one a competent developer could plausibly ship — the step explicitly rejects absurd alternatives, with the reason stated.
- The step's text states its division of labor with `plan-review-acceptance` and why the earlier placement catches something the later check structurally cannot.
- No new gate, severity scheme, or confidence score is introduced — the check only moves items between the two existing classifications.
- A criterion whose alternative *is* ruled out by the source stays `inferable`, with the check recorded as passed rather than omitted.

## Ambiguity Log

| Decision | Classification | Resolved By | Rationale / Answer |
|----------|---------------|-------------|-------------------|
| Placement between Step A and Step B | `inferable` | inference | The check's output is a classification, so it must run after inference is attempted and before the classification is recorded. Anywhere else and it is a separate pass with no consumer. |
| Exactly one alternative per criterion | `inferable` | inference | A tripwire, not an enumeration. Multiple alternatives per criterion turns a classification aid into an analysis phase and inflates the Ambiguity Log past readability. |
| Plausible rather than absurd alternatives | `inferable` | inference | Deliberate divergence from DeFOSPAM's Paul. Their findings are advisory and scored; ours flip a hard block. An absurd alternative produces a false block, and false blocks are how a gate gets ignored. |
| Not a duplicate of `plan-review-acceptance` | `inferable` | inference | Epic #2159 constraint. The two apply the same *test* at different *stages* against different *inputs* (source text vs. our restatement). Stated in the step text so the boundary survives future edits. |
| Recording the alternative in the Ambiguity Log | `inferable` | inference | The alternative is the question the human is being asked. Without it the log says an item blocks but not what decision is pending. |
| Passed checks recorded, not omitted | `inferable` | inference | An omitted check is indistinguishable from a skipped one. The log is the audit trail that makes "we asked before building" an artifact rather than an assertion. |

## Consistency Gate
- [x] Intent is unambiguous
- [x] Every behavior/goal maps to an acceptance criterion
- [x] Architecture constrains without over-engineering
- [x] Terminology consistent across artifacts
- [x] No contradictions between artifacts
- [x] Every gap/ambiguity finding is logged — inferable with rationale or resolved by human

### LINKED ISSUE #2179: refactor(specs): extract the persistence procedure into a loaded reference
## Intent Description

`skills/specs/SKILL.md` carried its entire persistence procedure inline — the file-vs-GitHub-issue branching, the body template, and the `/plan` auto-trigger rule. That was 1,093 of the file's 2,659 words: 41% of the skill, for a leaf step nothing above the Consistency Gate refers into.

Every sibling skill of comparable size already splits its procedure into `references/` (`plan/references/`, `build/references/`, `test-improve/references/`). `skills/specs/` was the outlier, and all four slices of this epic add to the same file.

This slice is **convention alignment, not urgency**. There is no documented per-skill token ceiling in this repo, and a peer skill (`cd-test-architecture`) is larger. Extraction also does not, by itself, relax the epic's ~40-line cap, which is a per-slice delta — so this slice ships the deterministic check that makes both that cap and a cumulative ceiling enforceable, rather than leaving them to author judgment.

Folded into epic #2159 at the operator's direction as a prerequisite for slices #2160–#2163.

## Architecture Specification

- **Move**, verbatim: `## Output` through `### Auto-trigger /plan` into `skills/specs/references/persistence.md`, heading levels flattened now that the sections are top-level in their own file. SKILL.md keeps a short `## Output` stub plus an explicit load pointer.
- **No content loss**: every non-blank source line must be present after the move.
- **Correct the recorded size** in `knowledge/agent-registry.md`, which still carried the pre-extraction figure. The canonical measurement is `scripts/measure_tokens.py`, not an approximation.
- **Add `scripts/specs_skill_delta.py`**, stdlib-only, with two independent checks: growth since a recorded baseline against the epic's ~40-line cap, and a cumulative ceiling against the pre-extraction size. The second exists because slices can each pass the per-slice cap while collectively re-consuming the headroom — a case a per-slice cap structurally cannot see.
- **Wire it into `scripts/ci-local.sh`.** A gate nothing invokes cannot fail, and CLAUDE.md is explicit that such a gate is worse than none.

## Acceptance Criteria

- [ ] The persistence procedure lives in `references/persistence.md`, reachable from SKILL.md by an explicit load pointer.
- [ ] Both persistence branches, the single fenced body template, and the auto-trigger rule survive the move intact.
- [ ] SKILL.md no longer inlines the body template or the branching.
- [ ] `knowledge/agent-registry.md`'s figure for the specs skill matches `measure_tokens.py --verify` with no drift.
- [ ] `scripts/specs_skill_delta.py --check` enforces both the since-baseline cap and the cumulative ceiling, and exits non-zero on either breach.
- [ ] Its baseline is stored at a path git can actually track.
- [ ] `baseline` and `--check` cannot be combined — recording a baseline from a breaching file would launder the breach into the new normal.
- [ ] The check is registered in `ci-local.sh`'s CHECKS roster.
- [ ] The gate is made to fail on purpose before being trusted.
- [ ] Every markdown reference still resolves.

## Notes

Filed after implementation to give the batch PR a number to close. The work is on `claude/nifty-albattani-82ujmh`.

bdfinst:
## Summary

- **`/specs` can now validate a requirements document someone else wrote.** It was an authoring loop only, so a stakeholder PRD or vendor RFP reached `/plan` unexamined — the first artifact any gate saw was already our restatement, with every ambiguity in the source silently resolved. Step 0 now picks a mode from the argument's shape, with no flag, so existing invocations are untouched.
- **Three gaps that were structurally invisible are now findings, not silences**: a completeness sweep for entities and concerns the spec never mentions, a glossary recording which domain terms a human has actually verified, and a predictability check that makes an `inferable` classification falsifiable at the point it is applied.
- **A prerequisite slice extracted the persistence procedure** (41% of `SKILL.md`'s words, a leaf step) into `references/persistence.md`, and added a deterministic size budget so the epic's "~40 lines per slice" constraint is enforced by a script rather than by eye.

Every slice routes into the **existing** Ambiguity Resolution Protocol and Ambiguity Log. No new gate, severity scheme, or confidence score anywhere — the deliberate rejection of DeFOSPAM's 1–10 scale, which silently drops findings below 7 where a binary classification drops nothing.

## Test Plan

- [x] Full pre-push suite green — 10,705 passed, 48 skipped
- [x] `bash scripts/ci-local.sh` green, including the new `chk_specs_skill_size` gate
- [x] `python3 scripts/check_md_references.py` — every reference resolves
- [x] `python3 scripts/measure_tokens.py --verify` — specs row at 0.0 drift
- [x] `python3 plugins/dev-team/hooks/lib/build_knowledge_index.py --check` — index fresh
- [x] New size gate made to fail on purpose before being trusted: +45 lines trips the per-slice cap alone, +85 trips both, a missing/corrupt/foreign baseline fails closed, and `baseline --check` is rejected
- [x] 116 new doc-guard assertions across six test files
- [ ] Reviewer: confirm the two deviations below are acceptable

## Decisions & assumptions

**Two sub-issues were implemented differently from their written specs, both approved at the plan gate:**

1. **#2161** names `knowledge/completeness-checklist.md`. It ships at `skills/specs/references/completeness-checklist.md` instead — `knowledge/` is this repo's convention for cross-agent shared content, `specs/SKILL.md` had zero `knowledge/` references, and this checklist serves `/specs` alone.
2. **#2162** says the Glossary goes in "both persistence templates… the two templates already mirror each other". There is only **one** fenced body template; the GitHub-issue branch composes its body by citing that template's section list. Implementing it literally would have created the second template — and therefore the drift — its own rationale warns against. The Glossary went into the single template, the issue branch's cited section list gained it, and `test_exactly_one_body_template_still_exists` guards against a second appearing.

**The size gate earned its place twice, both on the case a per-slice cap cannot see.** Slice 4 landed +25 lines — well under the 40-line cap — yet pushed `SKILL.md` past its pre-extraction size; the cumulative check caught it and the prose was re-cut into `references/glossary.md`, dropping it to +9. Slice 5 hit exactly 243, the `>=` boundary its unit test pins, and was trimmed to +8. Final size is 238 lines against a 243 ceiling.

**`references/` sits outside the bash-grant guard's coverage.** `tests/repo/test_skill_bash_grants_cover_invocations.py` globs `skills/*/SKILL.md` only. Moving the persistence procedure — which instructs `gh issue list`/`create`/`edit` — into a reference moves it out of that scanner's view. It passes today only because those invocations are inline code rather than fenced bash blocks; a future fenced block added to a reference would not be caught. Worth a follow-up, out of scope here.

**Pre-existing, found in passing, not fixed here:** `SKILL.md`'s Step 0 version check tells the agent to compare against "the current skill version (see frontmatter `version:`)", but the frontmatter has no `version:` field. Present in the shipped 13.1.2 plugin too.

## Review notes

Reviewed by `correctness-review`, `test-review`, `doc-review`, `structure-review`, and `spec-compliance-review` across per-slice checkpoints and a final backstop. Spec-compliance returned PASS on all five issues.

The correctness pass on the new script found four ways it could not actually gate anything — a gitignored baseline that made it enforceable only on the authoring machine, a `baseline --check` combination that recorded a breaching file as the new normal and exited 0, a provenance key written but never read, and no CI wiring at all. All four are fixed in `a493f8f`. The repo's own `measure_tokens.py --verify` separately blocked the first push and was right to: a words×4/3 token estimate was 20% low, and the test that used it now calls the canonical tool instead.

Closes #2179
Closes #2160
Closes #2161
Closes #2162
Closes #2163
Part of #2159

https://claude.ai/code/session_016uBnw1i52qEik2k9LSHa4n

---
_Generated by [Claude Code](https://claude.ai/code/session_016uBnw1i52qEik2k9LSHa4n)_


bdfinst:
## One action needed from a maintainer

`chk_specs_skill_size` is registered in `ci-local.sh`'s CHECKS array but **is not named by any CI job's `--only=` list**, so it currently runs only on a contributor machine via `pre-push`. That is exactly the condition `ci-local.sh`'s own comment block warns about — "an always-run check that no CI job names is a gate only a correctly-configured contributor machine enforces" — and the same failure mode that comment records for `ruff`.

The fix is one line in `.github/workflows/plugin-tests.yml:130`, adding `chk_specs_skill_size` to the structural-gates job's list:

```
--only=chk_md_references,chk_skills_index,chk_rules_vs_prompts,chk_hook_units,chk_sa_mcp_tools,chk_specs_skill_size
```

I made that change and the push was rejected: my token has no `workflow` scope. Rather than leave the gap silent, I backed the file out and documented it in `ci-local.sh`'s exception comment, marked explicitly as a gap rather than a deliberate exception.

Partial coverage in the meantime: the **cumulative** ceiling is incidentally checked in CI, because `tests/skills/test_specs_persistence_reference.py` re-asserts it and `tests/skills` runs under `chk_hook_units`, which the structural-gates job does name. The **since-baseline delta** — the reason the committed baseline file exists at all — runs nowhere in CI.

## Review round 2

The backstop found three defects beyond the first round, all fixed in `0176f63`:

- **Duplicate `## Step 0`**, mine — the new mode selector landed directly above the pre-existing "Step 0 — Existing-spec version check", so the file had two Step 0s and the step sequence stopped being a sequence. Selector keeps Step 0; version check becomes Step 1.
- **The baseline guard was incomplete.** Rejecting `baseline --check` blocked one spelling; plain `baseline` — the command the docstring tells operators to run as a slice's final action — still recorded a file 73 lines over the ceiling and exited 0. The ceiling now gates recording as well as checking.
- **`SKILL.md` over-duplicated the completeness checklist's prose** relative to how the other four reference splits are drawn. It keeps the requirements; the reference keeps the reasoning.

Also fixed in passing: two `proceed to Step 1` pointers that dangled before this branch (there has never been a `## Step 1`). Renumbering forced the issue, and leaving them inside a section now titled Step 1 would have been worse.

Still unfixed and out of scope, noted for a follow-up: `SKILL.md`'s Step 1 version check tells the agent to compare against frontmatter `version:`, which does not exist in the frontmatter. Present in the shipped 13.1.2 plugin too.


---

